PR #11, opened 2016-08-03
Develop

pashap:
@bobtfish @sarguru @CheRuisiBesares @Supernomad

I am down for merging PR #7, and then updating develop with those changes/fixes. 

I would really like to make this repo our ns1-go-sdk, where what we have now would be moved to a subpackage `rest`. Then we could also include a cli here as well. And making subpackages would probably make it easier to separate the rest methods (`CreateZone`) from non-rest methods (`NewZone`).

Please lmk your thoughts, Im worried that Im causing too many breaking changes. If so, we could always include a readme stating this repo is deprecated, and a new repo exists in its' place.


monitoring_test.go
```diff
@@ -86,8 +86,8 @@ func TestUnmarshalMonitoringJobs(t *testing.T) {
 		t.Error("Status is not up")
 	}
 	r := j.Rules[0]
-	assert.Equal(t, r.Key, "rtt", "RTT rule key is wrong")
-	assert.Equal(t, r.Value.(float64), float64(100), "RTT rule value is wrong")
+	// assert.Equal(t, r.Key, "rtt", "RTT rule key is wrong")
```
Supernomad: Why are these guys commented out?

pashap: I wanted to remove any non std lib dependencies. This was the only one I saw(are there more?). I think we will be able to still write the tests without `testify/assert`, but I should jump on that now. Think we should leave it in?

Supernomad: nope that makes sense completely, if we can get the tests working with out it no need to add the dep :+1:

api.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+	return c
+}
+
+// Debug enables debug logging
+func (c *APIClient) Debug() {
+	c.debug = true
+}
+
+type APIClientOption func(*APIClient)
```
Supernomad: Mind walking me through this idea? not sure entirely why we are returning a func as opposed to just adding the `SetClient` func to the `*APIClient` directly. Or is this from the change over and just left as is?

pashap: I wasnt sure if this was the best/right way to handle optional parameters(by passing in closures that set the attribute on the client), but I liked it after reading these:

http://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis
https://github.com/digitalocean/godo/issues/84

pashap:
@bobtfish @sarguru I believe you have already stated that you are cool with creating a `ns1-go-sdk` repo where we can make all sorts of breaking changes. We can add a note at the top of this repos README stating that this repo has been deprecated in favor of that one. 

Does this sound okay? Probably the only way to ensure that we dont ruin anyones day by breaking the package.

monitoring_test.go
bobtfish: I'm a big fan of testify/assert, and I think the dependency (as it's a test dependency that doesn't get compiled into the final binary) is *absolutely worth it* for the clarity it adds to the code.

That's just my 2p however, I'm ok with you dropping it if you feel differently.

bobtfish:
Could we instead tag the current repository as v1, then just increment the major version with you merge all the breaking changes, making v2?

We can then change the README / docs to suggest using gopkg.in, and we can also continue to support any old users (in the same repository), and also have a nicer path for doing breaking changes in the future.

Does that sound like a good alternative approach?

sarguru:
👍  for using gopkg.in

monitoring_test.go
pashap: @bobtfish I see what you mean, that is a dope testing package the more I look into it. I may have gone a bit overboard on no external deps(this is the only one, and like you said, not in the binary). Lets keep it.

pashap:
@bobtfish @sarguru I like your alternative in using gopkg.in

Ill make the current master `v1`(using tags), and leave develop alone for now. We can play with more ideas on the develop branch until we feel comfortable with releasing a `v2`. Sorry for the delayed response.

pashap:
Im closing this PR and creating an issue in its place(shouldve done that in the first place)

PR #25, opened 2016-10-28
"Record.Meta" can have non-string values.  Add "RegionMeta.Weight".

cakoose:


cakoose:
@bobtfish or @sarguru: Any idea how I can get my change merged?  The test failure looks unrelated to my change.

record.go
```diff
@@ -16,17 +16,17 @@ type Filter struct {
...
-	Filters         []Filter          `json:"filters,omitempty"`
-	Ttl             int               `json:"ttl,omitempty"`
-	UseClientSubnet bool              `json:"use_client_subnet"`
-	Regions         map[string]Region `json:"regions,omitempty"`
+	Id              string                 `json:"id,omitempty"`
+	Zone            string                 `json:"zone,omitempty"`
+	Domain          string                 `json:"domain,omitempty"`
+	Type            string                 `json:"type,omitempty"`
+	Link            string                 `json:"link,omitempty"`
+	Meta            map[string]interface{} `json:"meta,omitempty"`
```
pashap, line 24: This may cause breaking changes for people if they dont properly type cast meta values to strings, no?

Otherwise, feel free to checkout how [v2 handles metadata](https://github.com/ns1/ns1-go/blob/v2/rest/model/data/meta.go#L21)

pashap:
@cakoose The build is fixed from: https://github.com/ns1/ns1-go/pull/29, sorry about that.

As far as merging, if @bobtfish / @sarguru give a thumbs up, im down to merge this. i dont want to break any scripts/terraform provider implementations

cakoose:
Background: I'm trying to use [bobtfish/terraform-nsone-provider](https://github.com/bobtfish/terraform-nsone-provider), which depends on this library.  I ran into issues because `low_watermark` caused a crash and `weight` was silently discarded.

But like you said, the `Record.Meta` is backwards incompatible, so I'm not sure what to do...
- Create a new version on gopkg.in?
- Update terraform-nsone-provider to use v2?  (Seems like a _ton_ of work.)

pashap:
Ahh, that makes sense. Thanks for giving the background. 

I should do the same for you. Atm, we have two PRs open with terraform(https://github.com/hashicorp/terraform/pull/10172 and https://github.com/hashicorp/terraform/pull/10058). Ironically, we are still trying to figure out how to properly model ns1 metadata in terraform. Ive mentioned [how our metadata works in the PRs](https://github.com/hashicorp/terraform/pull/10058#pullrequestreview-8225445), and am waiting for feedback from the community or a terraform core member. 

I want to say that we should have a provider implemented/merged pretty soon.

cakoose:
Oh dang, that pull request has much newer code than [bobtfish/terraform-nsone-provider](https://github.com/bobtfish/terraform-nsone-provider).  In particular, it's using v2.

I guess I should wait for that to land.

PR #31, opened 2017-03-01
Monitoring Job/Notify List Clean up

pashap:
Fixes a few bugs around monitoring jobs and adds job status history endpoint.


rest/model/monitor/job_test.go
```diff
@@ -63,13 +63,21 @@ func TestUnmarshalJobs(t *testing.T) {
...
+
+	if j.Status["global"].Since != 1389407609 {
 		t.Error("since has unexpected value")
 	}
-	if status.Status != "up" {
+	if j.Status["global"].Status != "up" {
 		t.Error("Status is not up")
 	}
+
+	if j.Status["sjc"].Since != 1389404014 {
```
cpg1111, line 74: Is sjc an edge case for some reason?

rest/_examples/monitor_history.go
```diff
@@ -0,0 +1,75 @@
...
+	"net/http"
+	"os"
+	"time"
+
+	api "gopkg.in/ns1/ns1-go.v2/rest"
+	"gopkg.in/ns1/ns1-go.v2/rest/model/monitor"
+)
+
+// Helper that initializes rest api client from environment variable.
+func initClient() *api.Client {
```
cpg1111: So this is a bit of an opinionated thing, but you could define client outside of the function and make this a `func init(){}`, that way it runs before main does without explicitly calling the function, it's a style thing more than anything else though, I'd say it's fine as is.

rest/model/monitor/job_test.go
pashap: The conditional could have been "lga" as well, its purpose is to test that each region has its own `monitor.Status` instance.

This test case would fail before, when `Job.Status` was defined as `Status map[string]Status` instead of `Status map[string]*Status`

rest/_examples/monitor_history.go
pashap: Oh, i like that a lot. going to update

cpg1111:
Assuming tests pass, LGTM

PR #34, opened 2017-03-21
Fixing fallthrough behaviour for error handlers

### LINKED ISSUE #33: Error handling is wrong
Was trying to debug updating a record and I found this pattern repeated frequently:
https://github.com/ns1/ns1-go/blob/v2/rest/record.go#L89

```
if err != nil {
		switch err.(type) {
		case *Error:
			switch err.(*Error).Message {
			case "zone not found":
				return resp, ErrZoneMissing
			case "record already exists":
				return resp, ErrRecordExists
			}
		default:
			return resp, err
		}
	}
```

The inner switch for when the type is `*Error` has no default. So if it is an `Error` with a different message, the whole thing falls through and I get a nil error returned from the function, which is incorrect. 

I'd submit a PR, but this seems to be a problem throughout the entire library, not just this one place.

captncraig:
This fixes #33 for me, where non-matched error messages would result in a nil error being returned. Could likely remove the type switch altogether or replace all of these with a central message->error type lookup managed by client.Do. But this solves my immediate problem.

### DIFF
--- a/rest/account_apikey.go
+++ b/rest/account_apikey.go
@@ -49,9 +49,9 @@ func (s *APIKeysService) Get(keyID string) (*account.APIKey, *http.Response, err
 			if err.(*Error).Message == "unknown api key" {
 				return nil, resp, ErrKeyMissing
 			}
-		default:
-			return nil, resp, err
+
 		}
+		return nil, resp, err
 	}
 
 	return &a, resp, nil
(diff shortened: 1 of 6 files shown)


pashap:
This is great, thanks for finding and fixing the bug. Im going to merge this and create another issue to do as you propose regarding a central error lookup in Client.Do.

pashap (approved):

PR #37, opened 2017-05-02
README sample code: really give up without NS1_APIKEY

fcelda:


pashap (approved):
Thanks for the catch/fix!

PR #40, opened 2017-12-06
Convert ZoneRecord Tier field to float

flopex:
Caught this when using the terraform-provider-ns1 for which some zone records returned from the API would have the field `Tier` set to `1.0`

```json: cannot unmarshal number 1.0 into Go struct field ZoneRecord.tier of type int```

This was also reported in hashicorp/terraform#16761

Checking the API response from `https://api.nsone.net/v1/zones/:zone` it returns mixed records with the field `Tier` either as an int or float. This seems to only appear on recently modified records, ~1month.

Testing done:

`go test`

### DIFF
--- a/rest/model/dns/zone.go
+++ b/rest/model/dns/zone.go
@@ -48,7 +48,7 @@ type ZoneRecord struct {
 	ID       string   `json:"id,omitempty"`
 	Link     string   `json:"link,omitempty"`
 	ShortAns []string `json:"short_answers,omitempty"`
-	Tier     int      `json:"tier,omitempty"`
+	Tier     float64  `json:"tier,omitempty"`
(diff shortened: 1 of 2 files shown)


flopex:
@pashap would you be able to review this PR?

sarguru:
👍

sarguru (approved):

cpg1111:
Hey @flopex, great catch.  My only suggestion is using `json.Number` instead, that way we have the flexibility to use the value as both `int64`and `float64`.

pashap:
@flopex im sorry for the late response, could we try the approach that @cpg1111 mentioned with `json.Number`.

flopex:
@pashap np, I've updated with the suggested change.

pashap:
Thank you @flopex

PR #41, opened 2017-12-12
adds functions to make meta terraform friendly

alistanis:
I am happy to move these functions elsewhere if need be. I was having trouble getting terraform to work reliably with structures, so strings and reflection it will have to be. It's weird though because it worked when reading the data in, just not when saving the state. It's an internal terraform problem that I can't fix unless I make some significant changes to terraform itself.

### DIFF
--- a/rest/model/data/meta.go
+++ b/rest/model/data/meta.go
@@ -1,12 +1,20 @@
 package data
 
+import (
+	"encoding/json"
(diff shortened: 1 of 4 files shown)


pashap (approved):
This looks great

cpg1111 (approved):
lgtm

PR #42, opened 2017-12-13
Make meta terraform friendly

alistanis:
I don't love this, but we need to make what we get from the api match terraform's internal state representation of these values.

boolean => 1 or 0
floats => shortest possible decimal format


cpg1111 (approved):
LGTM

PR #44, opened 2017-12-14
adds metadata validation

alistanis:
Reflect all the things!

Seriously though, if there are any questions on this one I'm happy to oblige - and I want to make sure the validations I'm doing are accurate to the business logic we should have.

### DIFF
--- a/rest/model/data/meta.go
+++ b/rest/model/data/meta.go
@@ -1,9 +1,12 @@
 package data
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
+	"net"
 	"reflect"
+	"sort"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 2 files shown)


rest/model/data/meta.go
```diff
@@ -180,10 +183,7 @@ func FormatInterface(i interface{}) string {
 		}
 		return strings.Join(slc, ",")
 	case FeedPtr:
-		data, err := json.Marshal(v)
-		if err != nil {
-			panic(err)
-		}
+		data, _ := json.Marshal(v)
```
cpg1111, line 183: Curious, why aren't we handling the error anymore?  I agree with removing the panic, but perhaps we could bubble up the error and return it instead?

cpg1111 (changes requested):
Just one question around error handling a json marshal. Looks good otherwise.

rest/model/data/meta.go
cpg1111: Spoke IRL

cpg1111 (approved):

pashap (approved):

PR #46, opened 2017-12-16
revert changes that added f to floats as strings and allow support fo…

alistanis:
…r integers/floats in specific cases. i am not thrilled about this.

Unfortunately we need to store data exactly as terraform itself interprets it, which is a bit of a drag.


cpg1111 (approved):

alistanis:
ty homie

PR #49, opened 2018-06-12
removing omitempty by customer request

alistanis:
By customer request, we're removing omitempty so that customers can remove filters applied to a record by sending either [] or null - @jamorency 

Implications are that if customers want to update a record, they must always send the full set of filters, or they will not be applied. We've spoken about other possible breaking changes and have decided that this is still worth pursuing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-- 1.6.2
+- 1.10.3
 go_import_path: gopkg.in/ns1/ns1-go.v2
 script: script/test
 notifications:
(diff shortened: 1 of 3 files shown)


fcelda (approved):
LGTM. The underscores in example_test.go are intentional?

pashap (approved):

larsx2 (commented):
Looks just fine. What is the reason of the convention change?

alistanis:
@fcelda Yeah, interestingly when you're writing a function to test a method on a type, underscores are the idiomatic way of doing that. Travis recently updated their tests to fail example/test code that doesn't match identifiers consistent with that. @larsx2 In regard to omitempty, customers want to be able to remove filters, and to do that with the api, we need to send ```"filters": []``` or ```"filters": null```

PR #50, opened 2018-06-28
fix some meta names

mioi:
sometimes `ToCamel()` won't convert the json version of a meta name to what the meta data model expects.


alistanis (approved):
Thanks @mioi, LGTM

PR #51, opened 2018-06-29
fix issue creating new ns1 record

mioi:
after PR #49 was merged, there was an issue creating new records. this should fix it.

The exact error I was seeing, via Terraform (using the latest ns1 terraform provider at `master`):

```
1 error(s) occurred:

* ns1_record.rec3: 1 error(s) occurred:

* ns1_record.rec3: PUT https://api.nsone.net/v1/zones/serverless.farm/test3.serverless.farm/A: 400 Input validation failed (Value None for field '<obj>.filters' is not of type array)
```


alistanis (approved):
Ah, I didn't realize our API didn't like ```"filters": null```. Good catch!

PR #54, opened 2018-08-22
Fix bug when "up" is a "feed" in Terraform

mioi:
If an answer's `meta` field's `up` value is not a bool but a feed, it causes a panic and crashes terraform.

related to this issue:
https://github.com/terraform-providers/terraform-provider-ns1/issues/31

Add new case where interface could be a `map[string]interface{}`.

Also, allow the `up` value to be something other than a bool, since it can be a feed.

Example Terraform:
```hcl
resource "ns1_record" "blerb" {
  domain = "blerb.serverless.farm"
  zone   = "serverless.farm"
  type   = "A"
  ttl    = 300

  answers = {
    answer = "10.20.20.10"
    meta = {
      up = "${jsonencode(map("feed", "5b7c8c10a632f600017a90fa"))}"
    }
  }
}
```

Since the value of `up` has to be a string, i'm using TF's `jsonencode()` and `map()` functions to generate json like `{"feed":"5b7c8c10a632f600017a90fa"}`. The current behavior with this terraform code is that it'll `apply` successfully but a subsequent `plan` will show:
```
  ~ ns1_record.blerb
      answers.0.meta.up: "0" => "{\"feed\":\"5b7c8c10a632f600017a90fa\"}"
```

With this fix, a subsequent `plan` will show "No changes."

### DIFF
--- a/rest/model/data/meta.go
+++ b/rest/model/data/meta.go
@@ -179,6 +179,9 @@ func FormatInterface(i interface{}) string {
 			slc = append(slc, s.(string))
 		}
 		return strings.Join(slc, ",")
+	case map[string]interface{}:
+		data, _ := json.Marshal(v)
+		return string(data)
 	case FeedPtr:
 		data, _ := json.Marshal(v)
 		return string(data)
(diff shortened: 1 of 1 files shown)


alistanis:
Hey @mioi, like was discussed by our customer success team, we certainly really appreciate these fixes, but we're closing this until v0.12 is released. That will hopefully solve a lot of problems that the provider already has, and then we can revisit what FeedPtrs look like, since they are planning on introducing dynamic attributes in v0.12. Thanks for your patience!

mburtless:
Noting that a similar fix was eventually merged in #63 to support this use case while we (continue) to wait for v0.12

PR #55, opened 2018-08-27
Use correct parameter names for monitoring jobs

JonasPf:
Changed parameters are:

auth -> authorization
connection_timeout -> connect_timeout

The parameters are documented as part of the API.
To view the full list run this curl:

curl -X GET -H 'X-NSONE-Key: XXXXXXX' https://api.nsone.net/v1/monitoring/jobtypes


pashap (approved):
Nice find, thanks for fixing this.

PR #62, opened 2019-03-22
Fix sending empty TSIG struct when creating secondary zone

mburtless:
Creating a secondary zone by configuring a zone using the `MakeSecondary()` method of the zone model currently fails on Create() with the following error:
```
400 Input validation failed (Value None for field '<obj>.secondary.tsig' is not of type object)
```
This is because the TSIG field of the ZoneSecondary struct is included in the PUT request, even if it is empty.  This hotfix resolves this issue by adding `omitempty` to the TSIG field.


supernomad (approved):
nice catch

PR #63, opened 2019-04-25
Hotfix/allow feed ptr meta var

### LINKED ISSUE #35: Add the ability to configure feeds for records up filter
### Terraform Version
Terraform v0.11.7

### Affected Resource(s)
Please list the resources as a list, for example:
- ns1_record
- ns1_datafeed

### Terraform Configuration Files
```hcl
resource "ns1_datafeed" "api" {
  name      = "${ns1_record.primary_api.domain}"
  source_id = "${module.notification_list.datasource_ns1_monitoring}"

  config = {
    jobid = "${ns1_monitoringjob.api.id}"
  }
}

resource "ns1_record" "api" {
  zone   = "${ns1_zone.example_com.zone}"
  domain = "api.${ns1_zone.example_com.zone}"
  type   = "CNAME"
  ttl    = 900

  answers = [
    {
      answer = "${ns1_record.primary_api.domain}"

      meta = {
        priority = 0
        up       = true
      }
    },
    {
      answer = "${ns1_record.secondary_api.domain}"

      meta = {
        priority = 1
        up       = true
      }
    },
  ]

  filters = {
    filter = "up"
  }

  filters = {
    filter = "priority"
  }
}
```

### Expected Behavior
This configuration works find but I'd like to be able configure it as the following
```hcl
  answers = [
    {
      answer = "${ns1_record.primary_api.domain}"

     feeds = [
                {
                    feed = "${ns1_datafeed.api.id}"
                    source = "${module.notification_list.datasource_ns1_monitoring}"
                }
            ],

      meta = {
        priority = 0
        up       = {
          feed = "${ns1_datafeed.api.id}"
         }
      }
    },
    {
      answer = "${ns1_record.secondary_api.domain}"

      meta = {
        priority = 1
        up       = true
      }
    },
  ]
```

### Actual Behavior
Maybe there is another option to get this. I'd like to hear that

### References
When adding the feeds manually from NS1 portal, im getting the same crash as GH-31 mentioned

Thanks

mburtless:
Allows raw json for a feed pointer to be used as a value for a metadata key, which in turn enables a feed pointer to be generated when the caller generating the input does not have the ability to create new FeedPtr objects.

This is required to resolve https://github.com/terraform-providers/terraform-provider-ns1/issues/35 and enable a workaround to use feed pointers as metadata values via Terraform.

This PR also disables default behavior of casting all values for the `up` metadata key that != 1 as `false`.  This enables using a feed pointer as a value for `up` and also prepares for Terraform 0.12, in which string conversions of boolean configuration values will no longer be automatically converted to `1` or `0` (See https://github.com/hashicorp/terraform/issues/19618 for more info).

This PR is based on a fork by @glennslaven, a big thanks is owed for his contribution!

### DIFF
--- a/rest/model/data/meta.go
+++ b/rest/model/data/meta.go
@@ -247,9 +247,9 @@ func MetaFromMap(m map[string]interface{}) *Meta {
 		if _, ok := mt.FieldByName(name); ok {
 			fv := mv.FieldByName(name)
 			if name == "Up" {
-				if v.(string) == "1" {
+				if v.(string) == "1" || strings.ToLower(v.(string)) == "true" {
 					fv.Set(reflect.ValueOf(true))
-				} else if v.(string) == "0" {
+				} else if v.(string) == "0" || strings.ToLower(v.(string)) == "false" {
 					fv.Set(reflect.ValueOf(false))
 				} else {
 					fv.Set(reflect.ValueOf(ParseType(v.(string))))
--- a/rest/model/data/meta.go
+++ b/rest/model/data/meta.go
@@ -182,10 +182,12 @@ func FormatInterface(i interface{}) string {
 	case map[string]interface{}:
 		// Required for Terraform workaround to allow users to submit raw json of feed pointer
 		// as value for metadata.  See https://github.com/terraform-providers/terraform-provider-ns1/issues/35
-		mapVal := i.(map[string]interface{})
-		feedPtr := FeedPtr{FeedID: mapVal["feed"].(string)}
-		data, _ := json.Marshal(feedPtr)
-		return string(data)
+		if val, ok := v["feed"].(string); ok {
+			feedPtr := FeedPtr{FeedID: val}
+			data, _ := json.Marshal(feedPtr)
+			return string(data)
+		}
+		panic(fmt.Sprintf("expected map to contain 'feed' key to marshal as feedPtr, got: %+v", v))
 	case FeedPtr:
 		data, _ := json.Marshal(v)
 		return string(data)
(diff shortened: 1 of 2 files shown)


ionigman:
Tested placing meta/feeds into answers and records. Both work fine. Removal also works fine.

rest/model/data/meta.go
```diff
@@ -179,6 +179,13 @@ func FormatInterface(i interface{}) string {
 			slc = append(slc, s.(string))
 		}
 		return strings.Join(slc, ",")
+	case map[string]interface{}:
+		// Required for Terraform workaround to allow users to submit raw json of feed pointer
+		// as value for metadata.  See https://github.com/terraform-providers/terraform-provider-ns1/issues/35
+		mapVal := i.(map[string]interface{})
```
supernomad: isn't this just the `v` value from above? or is there a reason you are "recasting" `i` here?

supernomad (commented):
quick question

rest/model/data/meta.go
```diff
@@ -179,6 +179,13 @@ func FormatInterface(i interface{}) string {
...
+		feedPtr := FeedPtr{FeedID: mapVal["feed"].(string)}
```
supernomad: do we need to ensure `mapVal["feed"]` actually exists before casting here? Or is it just assumed that the "feed" key will always exist?

rest/model/data/meta.go
mburtless: Good catch, no reason to recast

rest/model/data/meta.go
mburtless: It was assumed, but we should definitely ensure it exists as no other key is valid.  Rewrote the case so it will panic if a `map[string]interface{}` is passed with any key other than `feed`, as that's the only valid value for this use case.

supernomad (approved):
:rocket:

PR #64, opened 2019-07-03
Fixes validation of IPPrefixes and extends test coverage

mburtless:
* Adds validation for IPPrefixes passed as []string from Terraform
* Fixes error handling in validateCidr so error is returned on first bad CIDR detected in slice.  Previous iteration would drop error if bad CIDR wasn't last value in slice
* Adds test coverage for validation of []string values passed from Terraform

### DIFF
--- a/rest/model/data/meta.go
+++ b/rest/model/data/meta.go
@@ -291,15 +291,25 @@ func validateCidr(v reflect.Value) error {
 			return err
 		}
 	}
-	var last error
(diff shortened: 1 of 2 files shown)


tlloydaubry (approved):
I guess I can't approve these? Wamp

tlloydaubry (approved):

pgray (approved):

PR #65, opened 2019-07-29
Adds support to secondary zones for additional IPs

mburtless:
Adds support to secondary zones struct for `other_ips` and `other_ports`.  Adds test coverage for `MakeSecondary` convenience function.


tlloydaubry (approved):
Looks fine to me

PR #67, opened 2019-09-23
support DNSSEC, also set up go mod and add Makefile

rupa:
* Support toggling DNSSEC on zones

* Add go.mod and go.sum (using 1.12)
* Add Makefile for tests (but leaves scripts for now)
* Tabs -> spaces in string literals in zone_test.go
* Little bit of README pointers for local dev and tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,9 @@
+all:
+
+test:
+	go fmt
+	go vet ./...
+	golint ./...
+	go test ./...
+
+.PHONY: all test
--- a/README.md
+++ b/README.md
@@ -56,8 +56,15 @@ Contributing
 
 Contributions, ideas and criticisms are all welcome.
 
+Run tests:
+
+```
+make test
+```
+
+Local dev: use `go mod replace` in client code to point to local checkout of
+this repository.
+
 # LICENSE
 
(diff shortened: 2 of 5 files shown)


rest/model/dns/zone.go
```diff
@@ -37,6 +37,9 @@ type Zone struct {
 	Primary *ZonePrimary `json:"primary,omitempty"`
 	// Secondary contains info for slaving the zone to a primary dns server.
 	Secondary *ZoneSecondary `json:"secondary,omitempty"`
+
+	// Whether or not DNSSEC is enabled on the zone
+	DNSSEC bool `json:"dnssec"`
```
rupa: couldn't use `omitempty` on a bool - prevents us from setting it false. doesn't seem like a back-compatibility issue tho

rest/model/dns/zone.go
```diff
@@ -158,4 +161,5 @@ func (z *Zone) LinkTo(to string) {
 	z.Pool = ""
 	z.Secondary = nil
 	z.Link = &to
+	z.DNSSEC = false
```
rupa: seems correct going by how the UI works

PR #68, opened 2019-09-23
Use go mod and Makefile

rupa:
Add go.mod and go.sum (using 1.12)
Add Makefile for tests (but leaves scripts for now)
Little bit of README pointers for local dev and tests


go.mod
```diff
@@ -0,0 +1,10 @@
+module gopkg.in/ns1/ns1-go.v2
```
mburtless, line 1: When I run `go mod tidy` on a new clone of `master`, my `go.mod` only requires `github.com/stretchr/testify v1.4.0`.  Any idea why all those indirect requirements are in there or if they are needed?

mburtless (approved):
LGTM, might be some unnecessary indirect requirements in `go.mod` though?

PR #69, opened 2019-09-23
support toggling DNSSEC on zones

rupa:
and a bit of whitespace cleanup in string literals in zone_test.go

### DIFF
--- a/rest/model/dns/zone.go
+++ b/rest/model/dns/zone.go
@@ -38,8 +38,9 @@ type Zone struct {
 	// Secondary contains info for slaving the zone to a primary dns server.
 	Secondary *ZoneSecondary `json:"secondary,omitempty"`
 
-	// Whether or not DNSSEC is enabled on the zone
-	DNSSEC bool `json:"dnssec"`
+	// Whether or not DNSSEC is enabled on the zone. Note we use a pointer so
+	// leaving this unset will not change a previous setting.
+	DNSSEC *bool `json:"dnssec,omitempty"`
 }
 
 func (z Zone) String() string {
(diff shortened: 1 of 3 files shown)


rupa:
I did test error handling with DNSSEC not enabled on account, we do get good clear error message about that.

rest/model/dns/zone.go
```diff
@@ -37,6 +37,9 @@ type Zone struct {
 	Primary *ZonePrimary `json:"primary,omitempty"`
 	// Secondary contains info for slaving the zone to a primary dns server.
 	Secondary *ZoneSecondary `json:"secondary,omitempty"`
+
+	// Whether or not DNSSEC is enabled on the zone
+	DNSSEC bool `json:"dnssec"`
```
mburtless: This effectively makes `"dnssec": false` the default for all newly created zones via the SDK.  Not strictly a problem for the SDK, but could create issues downstream.  I.e. in Terraform the `resourceZoneUpdate` function in our provider constructs the body for the update request via the `NewZone` function in the SDK.

If the tf provider is updated to use this version of the SDK and is used to manage an existing tfstate on a zone with dnssec enabled, it will disable dnssec on the next update via TF (since no `dnssec` attribute will be included in the tf config and `"dnssec": false` will be included in the body by the sdk).

Maybe this can be worked around when adding the dnssec attribute to the tf provider.  Might be cleaner to implement `omitempty` on this attribute by changing the type to `*bool`?

rupa: Good catch and nice way of solving it. This should be all set now.

mburtless (approved):
LGTM

PR #71, opened 2019-10-03
INT-49: Avoid div/0 in RateLimit.WaitTimeRemaining

rupa:
when rl.Remaining is 0, return the full rl.Period

### DIFF
--- a/rest/client.go
+++ b/rest/client.go
@@ -248,7 +248,13 @@ func (rl RateLimit) WaitTime() time.Duration {
 
 // WaitTimeRemaining returns the time.Duration ratio of Period to Remaining
 func (rl RateLimit) WaitTimeRemaining() time.Duration {
-	return (time.Second * time.Duration(rl.Period)) / time.Duration(rl.Remaining)
+	var remaining int
+	if rl.Remaining == 0 {
+		remaining = 1
+	} else {
+		remaining = rl.Remaining
+	}
+	return (time.Second * time.Duration(rl.Period)) / time.Duration(remaining)
 }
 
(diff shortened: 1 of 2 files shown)


mburtless (approved):
🚀

rest/client.go
```diff
@@ -248,7 +248,13 @@ func (rl RateLimit) WaitTime() time.Duration {
...
 // WaitTimeRemaining returns the time.Duration ratio of Period to Remaining
 func (rl RateLimit) WaitTimeRemaining() time.Duration {
-	return (time.Second * time.Duration(rl.Period)) / time.Duration(rl.Remaining)
+	var remaining int
+	if rl.Remaining == 0 {
+		remaining = 1
+	} else {
+		remaining = rl.Remaining
+	}
+	return (time.Second * time.Duration(rl.Period)) / time.Duration(remaining)
```
rupa: hmm, another option that might be a smidge faster would just be to divide by `time.Duration(rl.Remaining + 1)`

mburtless: Thinking about this a little deeper, don't we just want to `return (time.Second * time.Duration(rl.Period))` if ~`rl.Remaining == 0`~ `rl.Remaining < 2`?

mburtless (changes requested):

rest/client.go
rupa: yeah that would be functionally the same, but code would look a little cleaner. The "always add 1" approach would be more elegant, and it works fine when i tested it a bit against my super-rate-limited account, but it might have a bit higher chance of managing to hit a 429 than this impl.

mburtless (approved):
👍

PR #72, opened 2019-10-03
INT-49: Add another rate limit strategy

rupa:
RateLimitStrategyConcurrent might help in some use cases. It takes a
parallelism arg, and when remaining tokens are less than or equal to
that number, sleeps for WaitTime * parallelism.


mburtless (approved):
🚀

Issue #38, opened 2017-09-26
Failure unmarshalling CAA record


siwyd:
I'm running into an issue unmarshalling a CAA record for my domain. I think the interface is expecting an array of all strings in the JSON response, but the response has an array of mixed types, the first being an integer and followed by strings.

This is the script used:
```
package main

import (
	"fmt"
	"log"
	"net/http"
	"os"
	"time"

	api "gopkg.in/ns1/ns1-go.v2/rest"
)

func main() {
	k := os.Getenv("NS1_APIKEY")
	if k == "" {
		fmt.Println("NS1_APIKEY environment variable is not set, giving up")
		os.Exit(1)
	}

	httpClient := &http.Client{Timeout: time.Second * 10}
	doer := api.Decorate(httpClient, api.Logging(log.New(os.Stdout, "", log.LstdFlags)))
	client := api.NewClient(doer, api.SetAPIKey(k))

	records, _, err := client.Records.Get("siwyd.com", "siwyd.com", "CAA")
	if err != nil {
		log.Fatal(err)
	}

	for _, a := range records.Answers {
		fmt.Println(a)
	}

}
```

Output:
```
2017/09/26 13:38:13 go-ns1/2.0.0: GET https://api.nsone.net/v1/zones/siwyd.com/siwyd.com/CAA
2017/09/26 13:38:14 json: cannot unmarshal number into Go struct field Answer.answer of type string
exit status 1
```

If I change "CAA" to "MX", I get the proper output:
```
2017/09/26 13:37:56 go-ns1/2.0.0: GET https://api.nsone.net/v1/zones/siwyd.com/siwyd.com/MX
10 mxb.mailgun.org
10 mxa.mailgun.org
```

This is the actual JSON response with curl (notice the integer in the answer array):
```
{
  "domain": "siwyd.com",
  "zone": "siwyd.com",
  "use_client_subnet": true,
  "answers": [
    {
      "answer": [
        0,
        "issuewild",
        ";"
      ],
      "meta": {},
      "id": "59a6e940c9c79d000120d3e3"
    },
    {
      "answer": [
        0,
        "issue",
        "letsencrypt.org"
      ],
      "meta": {},
      "id": "59a6e940c9c79d000120d3e4"
    }
  ],
  "id": "59a6e940c9c79d000120d3e5",
  "regions": {},
  "meta": {},
  "link": null,
  "filters": [],
  "ttl": 3600,
  "tier": 1,
  "type": "CAA",
  "networks": [
[...]

PR #73, opened 2019-10-08
CAA support

ChrisMcKee:
Related to 
* https://github.com/terraform-providers/terraform-provider-ns1/issues/10
* https://github.com/ns1/ns1-go/issues/38

Swaps out the use of string[] for the record data for empty interface to avoid parsing issues caused by the API returning variable string/int parsing.

* I've added a zone example to `rest/_examples/zones.go`
* Changed all record data types in `rest/model/dns/answer.go`
* Added new method `NewCAAAnswer(flag int, tag string, value string) *Answer `
* Created an integration test (could do with full coverage via integration test considering the source of the error is the way the api functions per type) for the CAA record in `rest/record_test.go`

I've added a build feature tag (integrationtest) to the integration test as it requires 2 env vars setting; I used this to validate the change against our account.

### DIFF
--- a/rest/_examples/zones.go
+++ b/rest/_examples/zones.go
@@ -142,6 +142,23 @@ func main() {
 		}
 	}
 
+	caaRec := dns.NewRecord(domain, "", "CAA")
+
+	caaRec.Answers = []*dns.Answer{
+		dns.NewCAAAnswer(0, "issue", "digicert.com"),
+		dns.NewCAAAnswer(0, "issue", "sectigo.com"),
+		dns.NewCAAAnswer(0, "iodef", "sectigo.com"),
+	}
+	_, err = client.Records.Create(caaRec)
+	if err != nil {
+		// Ignore if record already exists
+		if err != api.ErrRecordExists {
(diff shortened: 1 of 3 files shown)


mburtless:
Hi @ChrisMcKee - This is awesome work, thanks so much for the very thorough contribution!

While using the empty interface is a great solution for the underlying issue, I'm concerned that it would be a breaking change for the SDK.

For example, rebuilding the Terraform provider off of this branch will cause the Records resource to fail in any place that currently relies on `Rdata` to be a `string`.  It will also fail on calls to `NewAnswer()`, as Go cannot directly convert `[]string` to `[]interface` (https://golang.org/doc/faq#convert_slice_of_interface).

Writing a custom JSON unmarshaller for this struct may be a less disruptive change, as it will allow us to keep the current types and type safety, but can handle the conversion of `int` to `string` in `Rdata`.

I took a shot at an implementation in #74 .  Would love to hear your feedback on that.

ChrisMcKee:
closed in favour of #74

PR #75, opened 2019-10-09
INT-62: set secondary disabled on MakePrimary

rupa:
Setting to nil means the API gets no secondary key, and we can end up
with both primary and secondary enabled.


rupa:
hmm, this won't work as is due to some apid code that I don't think is quite right. we 500 if we set secondary disabled but haven't set it enabled before.

rupa:
Withdrawn due to API interaction

PR #76, opened 2019-10-18
WIP INT-3: follow Link headers for pagination

rupa:
Applies to zones list, and get (for records)

WIP: to see if this approach is OK
TODO: tests, sorting?

### DIFF
--- a/rest/client.go
+++ b/rest/client.go
@@ -142,7 +142,7 @@ func SetFollowPagination(shouldFollow bool) func(*Client) {
 }
 
 // Do satisfies the Doer interface.
-func (c *Client) Do(req *http.Request, v interface{}) (*http.Response, error) {
+func (c Client) Do(req *http.Request, v interface{}) (*http.Response, error) {
 	resp, err := c.httpClient.Do(req)
 	if err != nil {
 		return nil, err
@@ -172,7 +172,7 @@ type NextFunc func(v *interface{}, uri string) (*http.Response, error)
 
 // DoWithPagination Does, and follows Link headers for pagination. Preserves
 // and returns the _first_ http.Response (to match Do).
-func (c *Client) DoWithPagination(req *http.Request, v interface{}, f NextFunc) (*http.Response, error) {
+func (c Client) DoWithPagination(req *http.Request, v interface{}, f NextFunc) (*http.Response, error) {
 	resp, err := c.Do(req, v)
 	if err != nil {
 		return nil, err
--- a/rest/client.go
+++ b/rest/client.go
@@ -142,7 +142,7 @@ func SetFollowPagination(shouldFollow bool) func(*Client) {
 }
 
 // Do satisfies the Doer interface.
-func (c Client) Do(req *http.Request, v interface{}) (*http.Response, error) {
+func (c *Client) Do(req *http.Request, v interface{}) (*http.Response, error) {
 	resp, err := c.httpClient.Do(req)
 	if err != nil {
 		return nil, err
@@ -167,11 +167,12 @@ func (c Client) Do(req *http.Request, v interface{}) (*http.Response, error) {
 	return resp, err
 }
 
-// nextFunc knows how to get and parse additional info from uri into v.
-type nextFunc func(v *interface{}, uri string) (*http.Response, error)
+// NextFunc knows how to get and parse additional info from uri into v.
+type NextFunc func(v *interface{}, uri string) (*http.Response, error)
 
-// DoWithPagination Does, and follows Link headers for pagination.
-func (c Client) DoWithPagination(req *http.Request, v interface{}, f nextFunc) (*http.Response, error) {
+// DoWithPagination Does, and follows Link headers for pagination. Preserves
+// and returns the _first_ http.Response (to match Do).
+func (c *Client) DoWithPagination(req *http.Request, v interface{}, f NextFunc) (*http.Response, error) {
 	resp, err := c.Do(req, v)
 	if err != nil {
 		return nil, err
--- a/rest/client.go
+++ b/rest/client.go
@@ -12,9 +12,11 @@ import (
 )
 
 const (
-	clientVersion    = "2.0.0"
-	defaultEndpoint  = "https://api.nsone.net/v1/"
-	defaultUserAgent = "go-ns1/" + clientVersion
+	clientVersion = "2.0.0"
+
+	defaultEndpoint               = "https://api.nsone.net/v1/"
+	defaultShouldFollowPagination = true
+	defaultUserAgent              = "go-ns1/" + clientVersion
 
 	headerAuth          = "X-NSONE-Key"
 	headerRateLimit     = "X-Ratelimit-Limit"
@@ -81,7 +83,7 @@ func NewClient(httpClient Doer, options ...func(*Client)) *Client {
 		Endpoint:         endpoint,
 		RateLimitFunc:    defaultRateLimitFunc,
 		UserAgent:        defaultUserAgent,
-		FollowPagination: true,
+		FollowPagination: defaultShouldFollowPagination,
 	}
 
 	c.common.client = c
@@ -163,6 +165,26 @@ func (c Client) Do(req *http.Request, v interface{}) (*http.Response, error) {
 	}
 
 	return resp, err
+}
+
+// nextFunc knows how to get and parse additional info from uri into v.
+type nextFunc func(v *interface{}, uri string) (*http.Response, error)
+
+// DoWithPagination Does, and follows Link headers for pagination.
+func (c Client) DoWithPagination(req *http.Request, v interface{}, f nextFunc) (*http.Response, error) {
+	resp, err := c.Do(req, v)
+	if err != nil {
+		return nil, err
+	}
+	nextURI := ParseLink(resp.Header.Get("Link")).Next()
+	for nextURI != "" {
+		nextResp, err := f(&v, nextURI)
+		if err != nil {
+			return nil, err
+		}
+		nextURI = ParseLink(nextResp.Header.Get("Link")).Next()
+	}
+	return resp, nil
 }
 
 // NewRequest constructs and returns a http.Request.
@@ -317,4 +339,16 @@ func SetStringParam(key, val string) func(*url.Values) {
 // SetIntParam sets a url integer query param given the parameters name.
 func SetIntParam(key string, val int) func(*url.Values) {
 	return func(v *url.Values) { v.Set(key, strconv.Itoa(val)) }
+}
+
+func (c *Client) getURI(v interface{}, uri string) (*http.Response, error) {
+	req, err := c.NewRequest("GET", uri, nil)
+	if err != nil {
+		return nil, err
+	}
+	resp, err := c.Do(req, v)
+	if err != nil {
+		return nil, err
+	}
+	return resp, nil
 }
--- a/rest/client.go
+++ b/rest/client.go
@@ -46,6 +46,9 @@ type Client struct {
 	// Func to call after response is returned in Do
 	RateLimitFunc func(RateLimit)
 
+	// Whether the client should handle paginated responses automatically.
+	FollowPagination bool
+
 	// From the excellent github-go client.
 	common service // Reuse a single struct instead of allocating one for each service on the heap.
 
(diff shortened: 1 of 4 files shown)


rest/client.go
```diff
@@ -74,10 +77,11 @@ func NewClient(httpClient Doer, options ...func(*Client)) *Client {
...
 	c := &Client{
-		httpClient:    httpClient,
-		Endpoint:      endpoint,
-		RateLimitFunc: defaultRateLimitFunc,
-		UserAgent:     defaultUserAgent,
+		httpClient:       httpClient,
+		Endpoint:         endpoint,
+		RateLimitFunc:    defaultRateLimitFunc,
+		UserAgent:        defaultUserAgent,
+		FollowPagination: true,
```
mburtless: Should we use a `defaultShouldFollow` var to contain the default value, instead of embedding it here?  Makes it easier to read if all defaults are kept in the same place instead of mixing embedding values and vars IMO.

rest/zone.go
```diff
@@ -52,9 +64,55 @@ func (s *ZonesService) Get(zone string) (*dns.Zone, *http.Response, error) {
...
+				return nil, resp, err
+			}
+			nextURI = ParseLink(nextResp.Header.Get("Link")).Next()
+		}
+	}
+
 	return &z, resp, nil
 }
 
+func (s *ZonesService) nextZones(zl *[]*dns.Zone, uri string) (*http.Response, error) {
```
mburtless: Does `zl` need to be a pointer to a slice?  Slice values are effectively passed by reference in Go, since the value of a slice is just a pointer to an underlying array and two other ints [1].  Accordingly, passing a slice by value is very cheap and will simplify your code below.

[1] https://blog.golang.org/go-slices-usage-and-internals

rest/zone.go
```diff
@@ -52,9 +64,55 @@ func (s *ZonesService) Get(zone string) (*dns.Zone, *http.Response, error) {
...
+		return nil, err
+	}
+	tmpZl := []*dns.Zone{}
+	resp, err := s.client.Do(req, &tmpZl)
+	if err != nil {
+		return nil, err
+	}
+	for z := range tmpZl {
+		*zl = append(*zl, tmpZl[z])
+	}
```
mburtless: `append` allows you to append a slice to another slice without manual iteration by using `...` to expand the second argument.

i.e. `zl = append(zl, tmpZl...)`

rest/zone.go
```diff
@@ -52,9 +64,55 @@ func (s *ZonesService) Get(zone string) (*dns.Zone, *http.Response, error) {
...
+	var tmpZone dns.Zone
+	resp, err := s.client.Do(req, &tmpZone)
+	if err != nil {
+		return nil, err
+	}
+	// Aside from Records, the rest of the zone data is identical in the
+	// paginated response.
+	for r := range tmpZone.Records {
+		z.Records = append(z.Records, tmpZone.Records[r])
+	}
```
mburtless: Same deal, this could be simplified to `z.Records = append(z.Records, tmpZone.Records...)`

rest/headers.go
```diff
@@ -0,0 +1,89 @@
...
+	for _, l := range commaRegexp.Split(s, -1) {
+		linkMatches := linkRegexp.FindAllStringSubmatch(l, -1)
+
+		if len(linkMatches) == 0 {
+			return nil
+		}
+
+		pieces := linkMatches[0]
+
+		link := &Link{URI: pieces[1], Extra: map[string]string{}}
```
mburtless: IMO we should validate that we received a legit URI before returning it to protect against API returning bad values.

(i.e. https://play.golang.org/p/JngFarWPF2-, https://stackoverflow.com/questions/31480710/validate-url-with-standard-package-in-go).

rest/zone_test.go
```diff
@@ -0,0 +1,161 @@
...
+    "domain":"c.coolzone.cool",
+    "short_answers": ["3.3.3.3"],
+    "type":"A"
+  },
+  {
+    "domain":"d.coolzone.cool",
+    "short_answers": ["4.4.4.4"],
+    "type":"A"
+  }]
+}`
```
mburtless: Are package scoped vars necessary here.  Can these live inside `mockAPI`?  Alternatively, can `mockAPI` be generalized by taking it's configuration (i.e. `http.HandlerFunc`)  from it's caller so it can be reused in the future?

mburtless (changes requested):
Implementation makes sense IMO, a couple of notes about types and places where code could be simplified or have stronger validation.

One thing I'm seeing is that `nextRecords` and `nextZones` are very similar.  Seems like the only difference is the type of the temp variable used to hold the results of the request and where they get appended.

I think you could factor this out into a more DRY implementation that lives in `client` by creating a func (i.e. `getNext`) that accepts a pointer to a temp var to hold results and a URI and performs the actual GET.  `nextRecords` and `nextZones` could just handle setting up the temp var and performing the append.  Could be useful if we have to reuse this behaviour for other endpoints in the future.

rest/client.go
rupa: done

rest/zone.go
rupa: done

rest/zone.go
rupa: done

rest/zone.go
rupa: I couldn't get this to work, I think because we aren't returning `zl`, we're mutating it as a side effect.

mburtless: Doh, I totally overlooked we weren't returning `zl`.  You're totally right.

rest/zone_test.go
rupa: not necessary but seemed more readable than having them inside `mockApi` and not very dangerous. Definitely don't mind moving them.

I thought about generalizing `mockAPI` more, and thought about ways of organizing the data it uses better, but prefer to wait until we have some other use cases. The repo could use some more tests and I'm happy to work on some :) for now, I kinda prefer it dumb-n-hardcoded like this, it's a simple place to start when we add more tests n stuff

rest/client.go
```diff
@@ -156,6 +167,26 @@ func (c Client) Do(req *http.Request, v interface{}) (*http.Response, error) {
 	return resp, err
 }
 
+// nextFunc knows how to get and parse additional info from uri into v.
+type nextFunc func(v *interface{}, uri string) (*http.Response, error)
```
mburtless: Love it.  This solution feels very Go, like http.HandlerFunc.

mburtless (approved):

rest/client.go
```diff
@@ -74,10 +79,11 @@ func NewClient(httpClient Doer, options ...func(*Client)) *Client {
 	}
 
 	c := &Client{
-		httpClient:    httpClient,
-		Endpoint:      endpoint,
-		RateLimitFunc: defaultRateLimitFunc,
-		UserAgent:     defaultUserAgent,
+		httpClient:       httpClient,
```
kevin-cantwell, line 82: If you use `httpClient: http.DefaultClient` instead, you can drop the `httpClient Doer` argument and the default invocation of `NewClient` just becomes: `rest.NewClient()`

rest/client.go
```diff
@@ -156,6 +167,26 @@ func (c Client) Do(req *http.Request, v interface{}) (*http.Response, error) {
 	return resp, err
 }
 
+// nextFunc knows how to get and parse additional info from uri into v.
+type nextFunc func(v *interface{}, uri string) (*http.Response, error)
+
+// DoWithPagination Does, and follows Link headers for pagination.
+func (c Client) DoWithPagination(req *http.Request, v interface{}, f nextFunc) (*http.Response, error) {
```
kevin-cantwell: I recommend you use a pointer receiver here `func (c *Client) ...`

As it's currently set up, I don't think it would cause any problems, but you may find it helps avoid problems down the road if any of your fields become non-pointer structs.

rest/client.go
kevin-cantwell: Oops, resolved this by accident!

rest/client.go
kevin-cantwell: You have mixed exposed and unexposed types in this signature. I'd pick one and stick with it. However, is there any strong reason why this function should be exposed at all?

rest/zone.go
```diff
@@ -55,6 +65,32 @@ func (s *ZonesService) Get(zone string) (*dns.Zone, *http.Response, error) {
...
 }
 
+// gets and appends another list of zones to the passed list
+func (s *ZonesService) nextZones(v *interface{}, uri string) (*http.Response, error) {
+	tmpZl := []*dns.Zone{}
+	resp, err := s.client.getURI(&tmpZl, uri)
+	if err != nil {
+		return nil, err
+	}
+	zoneList := (*v).(*[]*dns.Zone)
```
kevin-cantwell: Oof. I think this is convoluted and brittle. One has to think too hard about what's happening to `v` when invoking this as a `nextFunc` interface. 

Consider changing the signature of your pagination function to accept a page handler type:
`func (c *Client) DoWithPagination(req *http.Request, page func(*http.Response) error) error`

This allows the caller to handle each response explicitly, eg:
```go
if err := s.client.DoWithPagination(req, func(resp *http.Response) error {
  z, err := parseZone(resp)
  if err != nil {
    return err
  }
  zl = append(zl, z)
  return nil
}); err != nil {
  return nil, err
}
```

rest/zone.go
```diff
@@ -21,7 +21,12 @@ func (s *ZonesService) List() ([]*dns.Zone, *http.Response, error) {
 	}
 
 	zl := []*dns.Zone{}
-	resp, err := s.client.Do(req, &zl)
+	var resp *http.Response
+	if s.client.FollowPagination == true {
+		resp, err = s.client.DoWithPagination(req, &zl, s.nextZones)
```
kevin-cantwell, line 26: I think it's a problem if this function `List()` returns the raw http.Response, especially wrt pagination. Which page is it getting? The first? Last? Is it useful when paging is enabled?

rest/zone.go
```diff
@@ -55,6 +65,32 @@ func (s *ZonesService) Get(zone string) (*dns.Zone, *http.Response, error) {
...
+}
+
+// gets and appends another set of records to the passed zone
+func (s *ZonesService) nextRecords(v *interface{}, uri string) (*http.Response, error) {
+	var tmpZone dns.Zone
+	resp, err := s.client.getURI(&tmpZone, uri)
+	if err != nil {
+		return nil, err
+	}
+	zone := (*v).(*dns.Zone)
```
kevin-cantwell: Any type of casting like this is a code smell in Go. At the very least, you should be type checking here:
```
zone, ok := (*v).(*dns.Zone)
if !ok {
  return nil, errors.New(...)
}
```

kevin-cantwell (commented):
Dropped a few comments, but one larger concern I'd like to bring up is the package name.

The name `rest` is not a very "Go" package name here. It's too broad. Are we exposing an implementation of the REST protocol in this package? If not, it should be `ns1rest` or `ns1api` or better yet, the public client should be exposed simply under the root package: `ns1.NewClient`.

See our RFC on this subject for more info: https://ns1inc.atlassian.net/browse/RFC-10

rest/client.go
rupa: done. huh, this guy copied off of `Do`, which is/was the odd one out. Seems like `Do` could/should also have a pointer receiver

rest/client.go
rupa: This is definitely better, but a breaking change that doesn't seem worth it at the moment.

rest/zone.go
rupa: yeah good point. We probably don't need to return a Response, and shouldn't. Again tho, probably something we don't want to change right away due to breaking things.

rest/zone.go
rupa: ✔️

rest/zone.go
kevin-cantwell: It's debatable, but I think the pattern AWS uses with all their api interfaces is pretty flexible and maintainable: 
```
func (s *AWSService) ServiceCallXYZ(input * XYZInput) (*XYZOutput, error) {
  // all the http logic, retries, and pagination, etc.
}
```

rest/client.go
rupa: good point. `Do` could probably also be unexposed. But it could be nice - from an SDK as toolkit perspective - to let users get at some of the guts if they want do do something fancy.

For consistency and least amount of change, inclined to keep `DoWithPagination` exposed and uppercase `NextFunc`

rest/zone.go
rupa: yeah, agreed. I don't think the httpResponse should get out of these funcs. I don't think anything uses it or needs it - but that change will have to wait for a breaking version. For now, `DoWithPagination` matches `Do` in that it returns the first response

rest/zone.go
rupa: yeah this is not the cleanest or most maintainable approach, but I think we wanna stick with it for now. Longer term I think we'd like to advocate/discuss some API changes that would allow for finer-grained pagination control for users, without exposing them to confusing implementation details.

PR #77, opened 2019-10-25
INBOX-344: make pagination match standard behavior a bit more

rupa:
On HTTP error, the Do handler returns the response as well as the
error, and people were using that. In lieu of a breaking change,
we emulate this behavior as closely as we can in DoWithPagination.

DoWithPagination:
  * Return the response as well as the error, on HTTP Error
  * We now clobber the old responses as we traverse pages. On HTTP
    error, the response will match the error. Without error, the
    response will be the _last_ page of results.

Tried to exercise this in unit tests. Since it's a bit subtle, it
would be nice if a test broke if we accidentally change behavior.

Also:
  * Refactored the little "mockAPI" stuff for more generic use
  * Add a litle unit test for the hearer parsing funcs

### DIFF
--- a/rest/client.go
+++ b/rest/client.go
@@ -141,7 +141,9 @@ func SetFollowPagination(shouldFollow bool) func(*Client) {
 	return func(c *Client) { c.FollowPagination = shouldFollow }
 }
 
-// Do satisfies the Doer interface.
+// Do satisfies the Doer interface. resp will be nil if a non-HTTP error
+// occurs, otherwise it is available for inspection when the error reflects a
+// non-2XX response.
 func (c Client) Do(req *http.Request, v interface{}) (*http.Response, error) {
 	resp, err := c.httpClient.Do(req)
 	if err != nil {
@@ -172,7 +174,8 @@ type NextFunc func(v *interface{}, uri string) (*http.Response, error)
 
 // DoWithPagination Does, and follows Link headers for pagination. The returned
 // Response is from the last URI visited - either the last page, or one that
-// responded with a non-2XX status.
+// responded with a non-2XX status. If a non-HTTP error occurs, resp will be
+// nil.
 func (c Client) DoWithPagination(req *http.Request, v interface{}, f NextFunc) (*http.Response, error) {
 	resp, err := c.Do(req, v)
 	if err != nil {
--- a/rest/client.go
+++ b/rest/client.go
@@ -170,20 +170,21 @@ func (c Client) Do(req *http.Request, v interface{}) (*http.Response, error) {
 // NextFunc knows how to get and parse additional info from uri into v.
 type NextFunc func(v *interface{}, uri string) (*http.Response, error)
 
-// DoWithPagination Does, and follows Link headers for pagination. Preserves
-// and returns the _first_ http.Response (to match Do).
+// DoWithPagination Does, and follows Link headers for pagination. The returned
+// Response is from the last URI visited - either the last page, or one that
+// responded with a non-2XX status.
 func (c Client) DoWithPagination(req *http.Request, v interface{}, f NextFunc) (*http.Response, error) {
 	resp, err := c.Do(req, v)
 	if err != nil {
-		return nil, err
+		return resp, err
 	}
 	nextURI := ParseLink(resp.Header.Get("Link")).Next()
 	for nextURI != "" {
-		nextResp, err := f(&v, nextURI)
+		resp, err = f(&v, nextURI)
 		if err != nil {
-			return nil, err
+			return resp, err
 		}
-		nextURI = ParseLink(nextResp.Header.Get("Link")).Next()
+		nextURI = ParseLink(resp.Header.Get("Link")).Next()
 	}
 	return resp, nil
 }
@@ -347,9 +348,7 @@ func (c *Client) getURI(v interface{}, uri string) (*http.Response, error) {
 	if err != nil {
 		return nil, err
 	}
-	resp, err := c.Do(req, v)
-	if err != nil {
-		return nil, err
-	}
-	return resp, nil
+	// For non-2XX responses, Do returns the response as well as an error, for
+	// other errs, resp will be nil. Caller's responsibility to sort that out.
+	return c.Do(req, v)
 }
--- a/rest/zone_test.go
+++ b/rest/zone_test.go
@@ -2,6 +2,7 @@ package rest
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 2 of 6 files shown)


rest/client.go
```diff
@@ -170,20 +170,21 @@ func (c Client) Do(req *http.Request, v interface{}) (*http.Response, error) {
 // NextFunc knows how to get and parse additional info from uri into v.
 type NextFunc func(v *interface{}, uri string) (*http.Response, error)
 
-// DoWithPagination Does, and follows Link headers for pagination. Preserves
-// and returns the _first_ http.Response (to match Do).
+// DoWithPagination Does, and follows Link headers for pagination. The returned
+// Response is from the last URI visited - either the last page, or one that
+// responded with a non-2XX status.
```
mburtless: I think it would be worth it to explicitly call out that `resp` may be `nil` (i.e. if `Do` has an error making the actual request) as this code hides that possibility.

I see you're calling that out in `getURI`, but I think we should make this more obvious here, where users are more likely to see it.

rest/zone_test.go
```diff
@@ -9,56 +10,62 @@ import (
...
+func TestZone_list_pagination(t *testing.T) {
 	// It should follow Links and gather all Zones into the list
-	testServer := mockAPI(t)
+	testServer := mockAPI(t, responseMap{"/zones": handleZoneList})
 	doer := testServer.Client()
 	client := NewClient(doer, SetEndpoint(testServer.URL))
 
-	zones, _, _ := client.Zones.List()
+	zones, resp, err := client.Zones.List()
+	assert.Nil(t, err)
```
mburtless: FYI in case you haven't seen it, there's `assert.NoError()` which is essentially the same, but just a little more explicit when read.  The `if assert.NoError(t, err) { //some other enclosed tests }` pattern may be more readable here.

rest/zone_test.go
```diff
@@ -80,9 +90,36 @@ func TestPagination_disabled(t *testing.T) {
 	}
 }
 
-func mockAPI(t *testing.T) *httptest.Server {
+func TestZone_http_errors(t *testing.T) {
```
mburtless: I think one failure scenario is untested here: when `c.httpClient.Do(req)` in `Do` returns `err`.  This should return a nil `resp`.  Not a HTTP error, of course, so maybe it belongs in it's own test.

mburtless (changes requested):
Looks good, only tiny thing would be that I think we should just be more explicit about calling out and testing the one scenario when resp may be `nil`, since that's no longer obvious from `DoWithPagination()`.

Also, it would be great to ~use `assert.Equalf()` and~ (forgot `Equal` supports this out of the box) get `msg`s added to the assert statements, since these test funcs have multiple tests enclosed and I don't think some of what we are testing would be very clear on failure.  I think it's totally fine if we want to tackle that in another PR though.

rest/client.go
rupa: done, and noted in docstring for `Do` as well

rest/zone_test.go
rupa: nice, done

rest/zone_test.go
rupa: yeah, that's what I was having trouble thinking of how to test. Agree it should be a separate test. I came up with something, lets see if it feels reasonable to you.

mburtless (changes requested):
I think what you have is the correct solution for testing `Zones.List()`.

I failed at mentioning this earlier, but after thinking about this it occurs to me that we probably also want unit test coverage of `DoWithPagination()` and `getURI()` directly in `client_test.go`, right?  The coverage now seems closer to integration tests of `Zones.List()` and `Zones.Get()`.

Edit - One other piece of feedback that just occurred to me:  the naming scheme for these test functions seems strange to me.  I think what you typically see is `TestMyFunc` or `TestMyPackage_MyFuncSpecificBehaviour`, but it looks like the code switches to snake case for the piece that denotes the function and behavior we are testing.  Mixing case in function names seems like an antipattern.

I think the test functions that test multiple functions should also be broken up.  Functions and tests are usually one to one, in that a single test tests a single function.

rest/zone_test.go
```diff
@@ -9,80 +11,148 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestPagination_zoneList(t *testing.T) {
+func TestZone_list_pagination(t *testing.T) {
```
mburtless: Snake case function names are weird.  Should probably be named `TestZone_ListPagination`

rest/zone_test.go
```diff
@@ -9,80 +11,148 @@ import (
...
-	assert.Equal(t, 4, len(zones))
+	assert.Equal(t, 4, len(zones), "found wrong number of zones")
 	for idx := range zones {
-		assert.Equal(t, expected[idx], zones[idx].Zone)
+		assert.Equal(t, expected[idx], zones[idx].Zone, "zone.Zone does not match")
 	}
 }
 
-func TestPagination_zoneRecords(t *testing.T) {
+func TestZone_get_pagination(t *testing.T) {
```
mburtless: Should probably be named `TestZone_GetPagination`

rest/zone_test.go
```diff
@@ -9,80 +11,148 @@ import (
...
+	assert.Equal(t, "coolzone.cool", zone.Zone, "zone.Zone does not match")
+	assert.Equal(t, 4, len(zone.Records), "found wrong number of records")
 	for idx := range zone.Records {
 		assert.Equal(t, expected[idx],  zone.Records[idx].Domain)
 	}
+
 }
 
-func TestPagination_disabled(t *testing.T) {
+func TestZone_pagination_disabled(t *testing.T) {
```
mburtless: Should be split into two functions and named `TestZone_ListPaginationDisabled` and `TestZone_GetPaginationDisabled`

rest/zone_test.go
```diff
@@ -9,80 +11,148 @@ import (
...
 	for idx := range zone.Records {
-		assert.Equal(t, expected[idx],  zone.Records[idx].Domain)
+		assert.Equal(
+			t, expected[idx],  zone.Records[idx].Domain, "record.Domain does not match",
+		)
 	}
 }
 
-func mockAPI(t *testing.T) *httptest.Server {
+func TestZone_http_errors(t *testing.T) {
```
mburtless: Should be split into two functions and named something like `TestZone_ListWithHTTPErrors` and `TestZone_GetWithHTTPErrors`

mburtless (approved):
LGTM 🚀 

I'll just note that some of these tests seem like a good fit for table driven test pattern, but seems like a low priority to compared to getting this fix in.

PR #78, opened 2019-10-29
INBOX-344: force HTTPS when following Link headers

rupa:
When the Client endpoint is set to an HTTPS URL, rewrite any HTTP Link
URL to HTTPS.

### DIFF
--- a/rest/headers.go
+++ b/rest/headers.go
@@ -38,7 +38,7 @@ func (l Links) Next() string {
 
 // ParseLink parses a Link header value into a Links, mainly cribbed from:
 // https://github.com/peterhellberg/link/blob/master/link.go
-func ParseLink(s string) Links {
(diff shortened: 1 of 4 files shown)


rest/headers.go
```diff
@@ -38,7 +38,7 @@ func (l Links) Next() string {
 
 // ParseLink parses a Link header value into a Links, mainly cribbed from:
 // https://github.com/peterhellberg/link/blob/master/link.go
-func ParseLink(s string) Links {
+func ParseLink(s string, forceHTTPS bool) Links {
```
mburtless, line 42: Could use an updated description that describes the effect of `forceHTTPS`

mburtless (approved):
LGTM. I think we should document the new param in `ParseLink` though, as that's exported.

Issue #79, opened 2019-11-16
NS1 record to validate aws_acm_certificate fails - trailing dot


sam-caldwell:
This issue is submitted to tie this project to the reported issue in the ns1 terraform provider where the initial problem was identified.  See [Initial Issue Report](https://github.com/terraform-providers/terraform-provider-ns1/issues/38).

1. The issue is confirmed by multiple sources.
2. The issue was originally found in November 2018 and confirmed by the reporter of this issue in November 2019.
3. Investigation shows root cause of the issue and a possible solution.

This issue is being reported to associate with a pull request shortly and is intended to close the loop between the two projects.

sam-caldwell:
The following pull request is submitted to address this issue: https://github.com/ns1/ns1-go/pull/80.

Please let me know if there are any further questions.

Zach-Johnson:
A separate PR was opened and merged to address this here: https://github.com/terraform-providers/terraform-provider-ns1/pull/97

Zach-Johnson:
The PR mentioned above has been merged and released so I'm closing this issue.

PR #80, opened 2019-11-16
issue-79: strip leading and trailing dot characters 

sam-caldwell:
**Problem:**
As reported in _ns1-go_ [issue #79](https://github.com/ns1/ns1-go/issues/79) and _terraform-provider-ns1_ [issue #38](https://github.com/terraform-providers/terraform-provider-ns1/issues/38), certain use cases exist where a preceding or trailing dot character may exist in a domain or zone.  When passed to _ns1/ns1-go_, the _NewRecord()_ function will concatenate the domain and zone without stripping the dot character and result in something like domain..zone.  This results in an HTTP/400 error from NS1 API, since it cannot handle trailing dot characters.

For example, Amazon AWS Certificate Manager (ACM) generates DNS validation records which end in a trailing dot.  Because of the bug addressed in this ticket, _terraform-provider-ns1_ generates calls _NewRecord()_ with the dot-terminated domain and zone, and _NewRecord()_ produces a malformed request.

**Solution:**
We have added a function _removeEnclosingDots()_ which is called by _NewRecord()_ to strip any preceding or trailing dot characters prior to evaluating _.HasSuffix()_.

### DIFF
--- a/rest/model/dns/record.go
+++ b/rest/model/dns/record.go
@@ -35,9 +35,12 @@ func (r Record) String() string {
 // NewRecord takes a zone, domain and record type t and creates a *Record with
 // UseClientSubnet: true & empty Answers.
 func NewRecord(zone string, domain string, t string) *Record {
+	domain = removeEnclosingDots(domain)
+	zone = removeEnclosingDots(zone)
 	if !strings.HasSuffix(domain, zone) {
 		domain = fmt.Sprintf("%s.%s", domain, zone)
 	}
+
 	return &Record{
 		Meta:    &data.Meta{},
 		Zone:    zone,
@@ -74,4 +77,20 @@ func (r *Record) AddFilter(fil *filter.Filter) {
 	}
 
 	r.Filters = append(r.Filters, fil)
+}
+
+// Remove any dot (.) character prefix or suffix of the given string.
+// This is useful because dns names often end in '.' characters to signify
+// the root of the DNS tree.  But NS1 API does not support this.
+//
+// In other cases a domain or zone may be passed in with a preceding dot (.)
+// character which would likewise lead the system to fail.
+func removeEnclosingDots(inputString string) string {
+	if strings.HasSuffix(inputString, ".") {
+		inputString = inputString[0 : len(inputString)-1]
+	}
+	if strings.HasPrefix(inputString,"."){
+		inputString = inputString[1:len(inputString)]
+	}
+	return inputString
 }
--- a/rest/model/dns/record.go
+++ b/rest/model/dns/record.go
@@ -35,8 +35,8 @@ func (r Record) String() string {
 // NewRecord takes a zone, domain and record type t and creates a *Record with
 // UseClientSubnet: true & empty Answers.
 func NewRecord(zone string, domain string, t string) *Record {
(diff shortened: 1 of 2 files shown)


rupa:
Hey, thanks for the issue and PR! I'm not sure about this one. If we use this change in terraform provider and give some "dotted" up zone/domains it's not quite a drop-in - the record gets created correctly but terraform sees things as dirty in subsequent runs, as the original "dotted" strings are what's stored in state. Haven't messed with it yet, but I'm similarly not sure if this would play nicely with changing/deleting the record. To be fair, you might have a corresponding change for terraform-provider in the wings, but even then, I'm not sure this is the best approach.

A couple questions/comments on my mind that could help make sure we're on the same page and get this solved in a way we're all OK with:
* is your interest here tightly coupled with usage in terraform-provider or will this bite you separately just using the SDK?
* Our API looks to be pretty lenient with accepting FQDNs with trailing dots and normalizing them in the response. That doesn't help with leading dots, of course (curious if you are actually seeing those or are you handling them "since we're here"?). It also leaves TF thinking things are dirty.
* What would be simplest in terms of terraform would be doing this dot-replacement in the config language. Unfortunately, user defined functions aren't supported, but we do have a builtin `replace` function that seems to do the trick, if not in the most pleasant way possible. Something like this worked for me - stripped what it should and not what it shouldn't, allowed for using variables from other resources, and kept terraform "clean":

```
resource "ns1_record" "burpingcat_cname" {
  zone   = replace("${ns1_zone.burpingcat.zone}", "/(^\\.)|(\\.$)/", "")
  domain = replace(".foo.burping.cat.", "/(^\\.)|(\\.$)/", "")
  type   = "CNAME"

  answers {
    answer = "bar.${ns1_zone.burpingcat.zone}"
  }
}
```

Right now I'm thinking maybe just showing an example of the above in our terraform provider docs would be helpful and sufficient. For the SDK, I wouldn't mind merging just the `RemoveEnclosingDots` function as a convenience (without actually using it in `NewRecord`) but i could see an argument that it's out of scope for SDK, and not complex enough to warrant including anyway.

Anyway, thanks again, hope to hear your thoughts and opinions, happy to discuss/argue/whatever about this further!

sam-caldwell:
Calling replace() as a workaround really hurts for NS1 customers.   It cost me a day of troubleshooting, given that everything was working for one (non-NS1) environment but the same terraform broke when I went to the environment that uses NS1.

I agree that we need a change to terraform as well.  Let me look at that side again, as this is where I started.

I had tried to create a developer account to do acceptance tests, but the system doesn't seem to have provisioned that account, so much of my work is based on what I can do without an NS1 account in my free time.

rupa:
I was poking around getting together a response and it ended up easier to open a PR than to say words here. This approach, using `StateFunc` seems to work well https://github.com/terraform-providers/terraform-provider-ns1/pull/94 Does that seem like it would work for your use case?

regarding AccTests, no worries if you can't run 'em yet. In the meantime, either I'll get 'em written, or you can take a stab at em and i'll run/edit/tell you stuff.

rupa:
Going to close this in favor of the discussion over at https://github.com/terraform-providers/terraform-provider-ns1

PR #81, opened 2019-11-16
Git hooks fixes

sam-caldwell:
**Changes submitted:**
* Added `script/install-git-hooks`  to make it easy for new contributors to add the local git hooks.
* Updated documentation to support this change.
* Added fixes to git-hooks which allows the script to pass the shell check linter.

These are low-impact changes.

### DIFF
--- a/git-hooks/pre-commit
+++ b/git-hooks/pre-commit
@@ -13,10 +13,10 @@
 # This script does not handle file names that contain spaces.
 
 gofiles=$(git diff --cached --name-only --diff-filter=ACM | grep '.go$')
-[ -z "$gofiles" ] && exit 0
+[[ -z "$gofiles" ]] && exit 0
 
-unformatted=$(gofmt -l $gofiles)
-[ -z "$unformatted" ] && exit 0
+unformatted=$(gofmt -l "$gofiles")
+[[ -z "$unformatted" ]] && exit 0
 
 # Some files are not gofmt'd. Print message and fail.
 
--- a/git-hooks/pre-commit
+++ b/git-hooks/pre-commit
@@ -7,8 +7,8 @@
 
 # git gofmt pre-commit hook
 #
-# To use, store as .git/hooks/pre-commit inside your repository and make sure
-# it has execute permissions.
+# To use, execute script/install-git-hooks and all the hooks will
+# by installed via symlink.
 #
 # This script does not handle file names that contain spaces.
 
--- a/rest/model/dns/record.go
+++ b/rest/model/dns/record.go
@@ -35,12 +35,9 @@ func (r Record) String() string {
 // NewRecord takes a zone, domain and record type t and creates a *Record with
 // UseClientSubnet: true & empty Answers.
 func NewRecord(zone string, domain string, t string) *Record {
-	domain = removeEnclosingDots(domain)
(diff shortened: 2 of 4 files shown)


git-hooks/pre-commit
```diff
@@ -7,16 +7,16 @@
...
-# To use, store as .git/hooks/pre-commit inside your repository and make sure
-# it has execute permissions.
+# To use, execute script/install-git-hooks and all the hooks will
+# by installed via symlink.
 #
 # This script does not handle file names that contain spaces.
 
 gofiles=$(git diff --cached --name-only --diff-filter=ACM | grep '.go$')
-[ -z "$gofiles" ] && exit 0
+[[ -z "$gofiles" ]] && exit 0
```
rupa, line 16: these were OK with the single-`[`s I think. Would rather not reflexively change it unless you have some practical reason I'm not seein'

script/install-git-hooks
```diff
@@ -0,0 +1,11 @@
...
+
+(
+    cd """$(dirname "$0")""" || exit 1
+    cd .. || exit 1
+    for hook_source in git-hooks/*; do
+        hook_name=$(basename "$hook_source")
+        echo "Linking $hook_name from $hook_source to .git/hooks/$hook_name"
+        ln -sf "git-hooks/$hook_name" ".git/hooks/$hook_name"
+    done
+)
```
rupa, line 11: why the extra subprocess for this? since we're not sourcing the script we shouldn't end up `cd`'d anywhere

rest/model/dns/record.go
```diff
@@ -75,3 +78,19 @@ func (r *Record) AddFilter(fil *filter.Filter) {
...
+// character which would likewise lead the system to fail.
+func removeEnclosingDots(inputString string) string {
+	if strings.HasSuffix(inputString, ".") {
+		inputString = inputString[0 : len(inputString)-1]
+	}
+	if strings.HasPrefix(inputString,"."){
+		inputString = inputString[1:len(inputString)]
+	}
+	return inputString
+}
```
rupa: looks like these changes snuck in from the other PR. Want to think about the other one a bit, so if you wanna pull these out, we can get this stuff merged without waiting on it :)

rupa (changes requested):

rest/model/dns/record.go
sam-caldwell: Good point.  This PR was just low-hanging fruit from when I was working through linter errors. 
It's being fixed.

script/install-git-hooks
sam-caldwell: I've found it is generally safer since it means the user can execute it from other directions and not end up with arbitrary outcomes.

git-hooks/pre-commit
sam-caldwell: This is generally safer since no wordsplitting or glob expansion will be done for [[..]] versus using [..]  If you run shellcheck linter against this, it will not pass the linter.  I tend to use shellcheck to lint all of my shell code.

script/install-git-hooks
rupa: understand that you're `cd`ing to make building the links simpler, but wondering why we doing it in a subshell?

rupa (approved):

PR #83, opened 2019-11-20
Use `json.Number' for TTLs and fix Makefile

clickyotomy:
Hello,

I tried to run the example snippet from the README with a slight modification, and it doesn't seem to work because of JSON parse errors.

* Go Version
```
go version go1.13.4 darwin/amd64
```

* Snippet
```go
package main

import (
    "fmt"
    "net/http"
    "time"

    api "gopkg.in/ns1/ns1-go.v2/rest"
)

func main() {
    k := "{api-key}"
    httpClient := &http.Client{Timeout: time.Second * 10}
    client := api.NewClient(httpClient, api.SetAPIKey(k))
    zone, _, err := client.Zones.Get("example.com")

    if err != nil {
        fmt.Println(err)
        return
    }

    for _, r := range(zone.Records) {
        fmt.Println(r)
    }
}
```

* Output
```
json: cannot unmarshal number 60.0 into Go struct field Zone.ttl of type int
```

This PR changes the `.*TTL` fields from type `int` to `json.Number`, for which `.Float64`, `.Int64` methods can be used to get the actual values later on.

Also made a change to the Makefile to `go fmt` the files in the directory (please let me know if that should be in a different PR).


rupa:
Hi, thanks for the PR! Nice and welcome catch on the `go fmt` stuff.

I haven't looked at it in detail yet, but the type change looks like it could break existing uses, so it's unlikely to be accepted, at least right away. Looks like we should at least update the documentation though, and it seems likely we could improve things a bit and keep things very back-compatible.

If you want to separate out the `go fmt` changes though, I'll happily get that merged right away!

clickyotomy:
@rupa - Thanks for taking a look. Yes, it looks like this change might break a few things, including `terraform-provider-ns1`. I'll close this out, and maybe we can re-open this some other time (if need be).

Opened https://github.com/ns1/ns1-go/pull/85 for the Makefile stuff.

rupa:
@clickyotomy one thing to mention - I couldn't replicate this issue with my own setup - the TTLs should be ints. If you've found a repeatable way to get the TTL set/responding as e.g. `60.0` that could be something we'd want to fix :)

clickyotomy:
@rupa - Welp, it looks like there might be a problem with the response returned from the API itself. I checked the string response of the body in [`client.Do`](https://github.com/ns1/ns1-go/blob/v2/rest/client.go#L164-L166), before it was passed to [`json.Decoder.Decode`](https://golang.org/pkg/encoding/json/#Decoder.Decode) and it looks like the API returned the value `60.0` for `ttl`. The UI on the NS1 control panel just says "60" for the `SOA TTL` text box. Not sure why the response returned from the API has 60.0

Response (removed/modified some fields):
```json
{
   "nx_ttl":3600,
   "retry":7200,
   "zone":"example.com",
   "network_pools":[
      "pN"
   ],
   "serial":1234567890,
   "primary":{
      "enabled":false,
      "secondaries":[

      ]
   },
   "refresh":43200,
   "expiry":1209600,
   "dns_servers":[
      "foo.nsone.net",
      "bar.nsone.net",
   ],
   "records":["{removed}"],
   "meta":{
   },
   "primary_master":"foo.nsone.net",
   "ttl":60.0,
   "id":"some_id",
   "hostmaster":"hostmaster@nsone.net",
   "networks":[
      0
   ],
   "pool":"pN"
}
```

rupa:
yep, you're right. I did some sleuthing and my suspicion is that this record may have been created or updated before some of the current validation/coercion was in place. it might be worth contacting support so they can take a look at your account specifically and ensure everything is squared up going forward (or to nail down specifics if there's any issue with current behavior). I've also created a ticket for us to take a look at this (potential) issue in general, but I don't have a specific timeline on that.\

thanks again!

clickyotomy:
Will do, thanks for checking! :)

PR #84, opened 2019-11-20
Adding support for the URLFWD record

EvgenIn:
Added this as a solution for creating URLFWD records via TF 
PR#95


rupa:
I've got a PR up for handling URLFWD https://github.com/ns1/ns1-go/pull/86 and going to close this guy in favor of that. Feel free to take it for a test drive :) And I'll take care of the terraform-provider end once that's merged. (that one is just updating an enum, writing a test, and updating the SDK dependency).

Issue #82, opened 2019-11-19
Unable to create URLFWD records


ghost:
When creating a new URLFWD record the API endpoint returns a "Invalid redirect mode for URLFWD record" error.

The Answer struct defines the RData as a string array.  The API endpoint on the other hand appears to be expecting an integer for the Type, Forwarding Mode, and Query Forwarding values.

```
newRecord := dns.NewRecord("mydomain.com", "fwd.mydomain.com", "URLFWD")
newRecord.TTL = 3600
newRecord.AddAnswer(dns.NewAnswer([]string{"/", "https://www.google.com", "301", "2", "0"}))
```

Generates the following json request:

```
{
  "meta": {},
  "zone": "mydomain.com",
  "domain": "fwd.mydomain.com",
  "type": "URLFWD",
  "ttl": 3600,
  "answers": [
    {
      "meta": {},
      "answer": [
        "/",
        "https://www.google.com",
        "301",
        "2",
        "0"
      ]
    }
  ],
  "filters": []
}
```

Submitting the same request with curl, but changing the answer array as follows succeeds.

```
      "answer": [
        "/",
        "https://www.google.com",
        301,
        2,
        0
      ]
```

rupa:
Thanks for opening this issue. 

This sort of thing has come up before, so on the bright side some of the work done previously should help keep adding support pretty simple.

We'll want to add a `NewURLFWDAnswer` function that gets the values typed correctly for the API. We'll also need to ensure that the API response values get parsed back out correctly - there's a good chance that part will already work.

We'll get this work scheduled ASAP and I'll keep you posted.

rupa:
https://github.com/ns1/ns1-go/pull/86 should take care of this :)

PR #86, opened 2019-11-25
INT-90: support URLFWD records

rupa:
In previous cases, the API accepts strings, but has the correct value
types in the response. For URLFWD, the API requires correct types for
the aswer values on the inbound side as well.

* Add NewURLFWDAnswer convenience function. This allows passing args
  with the correct types (which are then coverted to all strings).
  NewAnswer can also be used, with the data passed as []string
* Custom MarshalJSON handles translating URLFWD's RData into API-speak.
  Response values for URLFWD are already handled with existing unmarshal
  logic.
* Started a table-driven test case for records. Made answer tests
  table-driven and added a URLFWD case

### DIFF
--- a/rest/model/dns/answer.go
+++ b/rest/model/dns/answer.go
@@ -142,3 +142,49 @@ func NewCAAAnswer(flag int, tag, value string) *Answer {
 		Rdata: []string{strconv.Itoa(flag), tag, value},
 	}
 }
+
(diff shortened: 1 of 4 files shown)


mburtless (approved):
LGTM.  The switch to table based tests is rad.

Noting that this closes #82

PR #88, opened 2019-12-17
fix a panic on url fwd on some invalid inputs

Zach-Johnson:


rupa (approved):
🌭

mburtless (approved):
LGTM 🚀 

Not sure what's causing Travis to fall over all of a sudden, but I see that `go get -u github.com/stretchr/objx` was never added to `bootstrap`? (Travis is still using `1.10`, so can't use `go mod download`)

Zach-Johnson:
Added that `go get` to the script and that seems to have fixed the travis issues @mburtless

PR #89, opened 2020-01-08
DDI Permissions

Zach-Johnson:
I tested this locally against managed and private and looks good to me. I also tested it against the private demo account, and it's failing because the demo is running an older version (2.1.1) and the permission for `security.manage_active_directory` isn't in the schema until I think 2.2.1. I'm thinking we should keep this as-is though, I don't want to remove it and then have to re-add it and maintain compatibility for both.

### DIFF
--- a/rest/model/account/permissions.go
+++ b/rest/model/account/permissions.go
@@ -6,6 +6,9 @@ type PermissionsMap struct {
 	Data       PermissionsData       `json:"data"`
 	Account    PermissionsAccount    `json:"account"`
 	Monitoring PermissionsMonitoring `json:"monitoring"`
+	Security   *PermissionsSecurity  `json:"security,omitempty"`
+	DHCP       *PermissionsDHCP      `json:"dhcp,omitempty"`
+	IPAM       *PermissionsIPAM      `json:"ipam,omitempty"`
 }
 
 // PermissionsDNS wraps a User's "permissions.dns" attribute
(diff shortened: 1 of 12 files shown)


rest/model/account/permissions.go
```diff
@@ -36,9 +41,30 @@ type PermissionsAccount struct {
 	ViewInvoices          bool `json:"view_invoices"`
 }
 
+// PermissionsSecurity wraps a User's "permissions.security" attribute.
+type PermissionsSecurity struct {
+	ManageGlobal2FA bool `json:"manage_global_2fa"`
+
+	// This field is only relevant for DDI and should not be set to true for managed.
+	ManageActiveDirectory bool `json:"manage_active_directory,omitempty"`
```
mburtless, line 49: Why not make this field's type `*bool` and get rid of `permissionsDDISecurity`?

My train of thought is, using a pointer would allow this to be omitted when using this struct against managed endpoint, but removes the need for a separate struct to maintain when hitting a DDI endpoint.  Side effect is that it would make this backwards compatible with `2.1.x`

mburtless (changes requested):
LGTM, just one question on `PermissionsSecurity`.

mburtless (approved):
LGTM 🌭

dwhittle-sfdc:
@Zach-Johnson @mburtless 
Using this updated version against v2.2.1 of the DDI product it is still not working for me:

`2020/01/13 12:08:06 ERROR: Unable to manage team configuration: PUT https://localhost/v1/account/teams: 400 Input validation failed (Unknown properties for field "permissions": "monitoring")`

The issue appears that the `monitoring` permissions are still being bundled into the JSON object and the API does not like that it is included since monitoring options are not currently available for the DDI product.

Here is a simplified version of what I am doing:

```go
t = &account.Team{Name: team.Name}
t.Permissions.DNS.ZonesAllow = []string{}
t.Permissions.DNS.ZonesDeny = []string{}
t.Permissions.DHCP = new(account.PermissionsDHCP)
t.Permissions.IPAM = new(account.PermissionsIPAM)
t.Permissions.Security = new(account.PermissionsSecurity)

// ... Note: I never touch t.Permissions.Monitoring

_, err = ns1.Teams.Create(t)
```

Zach-Johnson:
Hi @dwhittle-sfdc,

An additional `DDI` setting was added to the configuration for the client. You can set it like this:
```
client := api.NewClient(httpClient, api.SetAPIKey(apiKey), api.SetDDIAPI())
```
This should enable DDI support, and swap what permissions get bundled into the JSON object. If you hadn't already set that configuration option, can you try that?

We should probably clarify in the documentation that this needs to be set explicitly.

dwhittle-sfdc:
@Zach-Johnson That got it, at least for the `Team` object. I still have to use custom `User` and `APIKey` objects due to other fields I need that are not yet supported in the package (`invite_token` for `User`, and `ip_whitelist` for `APIKey`). I'll have to update my custom code to act like the package is now doing. Thanks

Zach-Johnson:
Glad that fixed it. Thanks for the heads up on some of those unsupported fields - I've added a ticket to our backlog to add support for missing fields and it should be prioritized soon.

PR #92, opened 2020-01-16
change ip whitelists to use a zero val on serialization

Zach-Johnson:
had to make this change because if no value is sent through, apid doesn't remove the whitelist, so Terraform was unable to delete ip whitelists properly.


mburtless (approved):
🥇

PR #96, opened 2020-02-03
serialize tsig enabled field when its value is false

### LINKED ISSUE #95: Unable to disable TSIG on a secondary zone
If TSIG is enabled on a secondary zone, it can't be disabled through this library because the field `zone.Secondary.TSIG.Enabled` is tagged with `json:"omitempty"`, so it will never be serialized to JSON if it's false.

This code will reproduce:
```
zone := dns.NewZone("secondary.test")
zone.MakeSecondary("1.2.3.4")
zone.Secondary.TSIG = &dns.TSIG{
	Enabled: true,
	Key:     "abc123",
	Name:    "test",
	Hash:    "hmac-sha512",
}
client.Zones.Create(zone)

zone.Secondary.TSIG.Enabled = false
client.Zones.Update(zone)
// zone still has TSIG enabled
```

acm1:
Fixes https://github.com/ns1/ns1-go/issues/95

### DIFF
--- a/rest/model/dns/zone.go
+++ b/rest/model/dns/zone.go
@@ -100,7 +100,7 @@ type TSIG struct {
 	Key string `json:"key,omitempty"`
 
 	// Whether TSIG is enabled for a secondary zone.
-	Enabled bool `json:"enabled,omitempty"`
+	Enabled bool `json:"enabled"`
 	// Which hashing algorithm
 	Hash string `json:"hash,omitempty"`
 	// Name of the TSIG key
--- a/rest/model/dns/zone_test.go
+++ b/rest/model/dns/zone_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"gopkg.in/ns1/ns1-go.v2/rest/model/data"
 )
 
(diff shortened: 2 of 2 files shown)


mburtless:
👋 Hi @acm1 - Thanks for submitting this PR and the issue.  This looks great!

Based on past experience with adjusting `omitempty` tags, we'll just want to get this tested against our Terraform provider prior to merging to see if this creates any unexpected behavior that we'll need to handle on that end.

I've created a ticket in our backlog to cover that and we'll circle back once that's complete.

mburtless:
After some internal discussion we're concerned about the potential impact of this change to existing code that utilizes this SDK.

Specifically, as this changes the behavior of the SDK to set `Enabled` to `False`  whenever a `TSIG` struct is defined, unless explicitly set to `True`, this could cause existing code to inadvertently disable TSIG.

We think that this field (and several other existing `bool` fields) would be better represented by a `*bool` type.  In conjunction with `omitempty` this gives the SDK the ability to more closely match the behavior of the API by representing all three acceptable states for this field in a request (`True`, `False` and ommited).

This would also be a breaking change that will need to be released in a new major version.  We're working on scoping what other changes we want to package into a major release and will update with a timeline soon.

If you feel inclined to rework this PR using the `*bool` type instead of `bool`, we'd love to review it and include it in the release.  Otherwise, we'll make sure to address this ourselves before we cut a new version.

PR #100, opened 2020-03-18
handle pulsar data in record answer meta

rupa:
We've mixed some terraform specific stuff in here, so we have to
handle it as slices (from API responses) or strings (from terraform)

Added a PulsarMeta type, for validation only, as it's probably a
breaking change to mess with the type on Meta.

### DIFF
--- a/rest/model/data/meta.go
+++ b/rest/model/data/meta.go
@@ -16,6 +16,13 @@ type FeedPtr struct {
 	FeedID string `json:"feed,omitempty"`
 }
 
+// PulsarMeta is currently only used for validation
+type PulsarMeta struct {
+	JobID     string  `json:"job_id,omitempty"`
+	Bias      string  `json:"bias,omitempty"`
+	A5MCutoff float64 `json:"a5m_cutoff,omitempty"`
+}
+
 // Meta contains information on an entity's metadata table. Metadata key/value
 // pairs are used by a record's filter pipeline during a dns query.
 // All values can be a feed id as well, indicating real-time updates of these values.
@@ -51,9 +58,8 @@ type Meta struct {
 	// float64 or FeedPtr.
 	LoadAvg interface{} `json:"loadavg,omitempty"`
 
-	// The Job ID of a Pulsar telemetry gathering job and routing granularities
-	// to associate with.
-	// string or FeedPtr.
+	// The Job ID of a Pulsar telemetry gathering job and associated metadata.
+	// list of PulsarMeta
 	Pulsar interface{} `json:"pulsar,omitempty"`
 
 	// GEOGRAPHICAL
@@ -177,6 +183,10 @@ func FormatInterface(i interface{}) string {
 		slc := make([]string, 0)
 		for _, s := range v {
 			switch ss := s.(type) {
+			// Pulsar
+			case map[string]interface{}:
+				data, _ := json.Marshal(v)
+				return string(data)
 			case string:
 				slc = append(slc, ss)
 			// The ASN field specifically is returned from the API as an integer,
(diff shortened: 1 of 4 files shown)


rest/model/data/meta.go
```diff
@@ -16,6 +16,13 @@ type FeedPtr struct {
 	FeedID string `json:"feed,omitempty"`
 }
 
+// PulsarMeta is currently only used for validation
+type PulsarMeta struct {
+	JobID     string  `json:"job_id,omitempty"`
+	Bias      string  `json:"bias,omitempty"`
+	A5MCutoff float64 `json:"a5m_cutoff,omitempty"`
+}
```
rupa, line 24: there's a "granularities" arg documented in API docs that send me down a bit of a rabbit hole before I figured out it's been disabled intentionally (PLS-61)

rest/model/data/meta.go
```diff
@@ -442,6 +458,33 @@ func validateNoteLength(v reflect.Value) error {
...
+		}
+		if err := json.Unmarshal(bs, &pulsars); err != nil {
+			return fmt.Errorf("pulsar: invalid value: `%v`", v.Interface())
+		}
+	} else if v.Kind() == reflect.String {
+		// String from terraform
+		if err := json.Unmarshal([]byte(v.String()), &pulsars); err != nil {
+			return fmt.Errorf("pulsar: invalid value: `%v`", v.String())
+		}
+	}
```
rupa, line 478: really wanted to not duplicate the Unmarshal calls but failed miserably at it :/

rest/model/data/meta.go
```diff
@@ -442,6 +458,33 @@ func validateNoteLength(v reflect.Value) error {
 	return nil
 }
 
+func validatePulsar(v reflect.Value) error {
+	var pulsars []*PulsarMeta
+
+	if v.Kind() == reflect.Slice {
```
mburtless: Could be cleaner as
```
switch v.Kind() {
    case reflect.Slice:
    ...
    case reflect.String:
    ....
```

rest/model/data/meta.go
```diff
@@ -442,6 +458,33 @@ func validateNoteLength(v reflect.Value) error {
...
+		if err != nil {
+			return fmt.Errorf("pulsar: unexpected value: `%v`", v.Interface())
+		}
+		if err := json.Unmarshal(bs, &pulsars); err != nil {
+			return fmt.Errorf("pulsar: invalid value: `%v`", v.Interface())
+		}
+	} else if v.Kind() == reflect.String {
+		// String from terraform
+		if err := json.Unmarshal([]byte(v.String()), &pulsars); err != nil {
+			return fmt.Errorf("pulsar: invalid value: `%v`", v.String())
```
mburtless: This validates that correct values were passed, but will ignore any fields not present in `PulsarMeta`.  That means a pulsar string passed from TF with invalid fields will pass validation and make it into the API request.  Ideally we should catch that here, no?

mburtless (dismissed):
🔥, maybe just needs some additional validation.

rest/model/data/meta.go
```diff
@@ -275,6 +285,12 @@ func MetaFromMap(m map[string]interface{}) *Meta {
 				} else {
 					fv.Set(reflect.ValueOf(ParseType(v.(string))))
 				}
+			case "Pulsar":
+				var pulsars []map[string]interface{}
+				err := json.Unmarshal([]byte(v.(string)), &pulsars)
```
Zach-Johnson: fwiw this could be `if err := .... ; err == nil`

rupa: ✔️

rest/model/data/meta.go
rupa: ✔️

rest/model/data/meta.go
rupa: point, but I'm inclined to leave it this way (for now at least) and just ensure we have the required field set. API will reject extra keys (but not in plan stage) and leaving some flexibility could be nice since these field options are not as in sync with documentation as they could be in general.

mburtless: 👍Totally, as is it fixes the bug and is obvi a huge improvement over what was there.

rupa:
rebased review changes into original commit so I could add the changelog/version bump without another PR and have a nice history (as long as I remember not to squashmerge)

mburtless (approved):
🌭 <sup>2</sup>

PR #101, opened 2020-05-24
Exported mock API service

iamgnat:
Hello, I'm submitting a mock package that users can use to mock the NS1 API service for their own unit testing. This PR covers the basic functionality that can be used to emulate any API request/response, but I plan to add more helper functions (see `mockns1/zone.go`) in later PRs to further simplify usage. I just didn't to do too much right now (both to make your review easier and so I didn't spend a lot of time on it if you don't approve :wink: ).

- `mockns1`
  - Initial add of the mock service
  - Added helper functions to simplify testing `api.Client.Zones` calls.
- `rest`
  - Refactored `zone_test.go` tests
    - Made use of the `mockns1` service
    - Moved the tests to the `rest_test` package to ensure exported functionality is verified.
    - Refactored from individual test functions to using the `t.Run()` sub-testing pattern.
    - Utilized `testify/require` instead of `testify/assert` as it stops the (sub)test execution rather than allowing it to continue.
    - Added tests for `Create()`, `Update()`, and `Delete()`.
  - Removed superfluous `== true` comparisons
    - Maybe consider switching to [golangci-lint](https://github.com/golangci/golangci-lint) over `go vet` and `golint`?

From `mockns1/example_test.go`:

```go
// Setup the mock service
mock, doer, err := mockns1.New(t)
require.Nil(t, err)

defer mock.Shutdown()

// Create your NS1 client and configure it for the mock service
ns1 := api.NewClient(doer, api.SetAPIKey("apikey"))
ns1.Endpoint, _ = url.Parse("https://" + mock.Address + "/v1/")

// Add your test case (zone list in this example)
require.Nil(t, mock.AddTestCase(http.MethodGet, "zones", http.StatusOK, nil, nil, "",
	[]*dns.Zone{{Zone: "foo.bar"}}))

// Perform your tests
zones, _, err := ns1.Zones.List()
require.Nil(t, err)
require.Equal(t, 1, len(zones))
require.Equal(t, "foo.bar", zones[0].Zone)
```

```
$ make test
go fmt ./...
go vet ./...
golint ./...
go test ./...
?   	gopkg.in/ns1/ns1-go.v2	[no test files]
ok  	gopkg.in/ns1/ns1-go.v2/mockns1	(cached)
ok  	gopkg.in/ns1/ns1-go.v2/rest	(cached)
?   	gopkg.in/ns1/ns1-go.v2/rest/model	[no test files]
ok  	gopkg.in/ns1/ns1-go.v2/rest/model/account	(cached)
ok  	gopkg.in/ns1/ns1-go.v2/rest/model/data	(cached)
ok  	gopkg.in/ns1/ns1-go.v2/rest/model/dns	(cached)
?   	gopkg.in/ns1/ns1-go.v2/rest/model/filter	[no test files]
ok  	gopkg.in/ns1/ns1-go.v2/rest/model/monitor	(cached)
```

### DIFF
--- a/mockns1/example_test.go
+++ b/mockns1/example_test.go
@@ -1,0 +1,38 @@
+package mockns1_test
+
+import (
+	"net/http"
+	"net/url"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"gopkg.in/ns1/ns1-go.v2/mockns1"
+	api "gopkg.in/ns1/ns1-go.v2/rest"
+	"gopkg.in/ns1/ns1-go.v2/rest/model/dns"
+)
+
+func Example() {
+	t := new(testing.T)
+
+	// Setup the mock service
+	mock, doer, err := mockns1.New(t)
+	require.Nil(t, err)
+
+	defer mock.Shutdown()
+
+	// Create your NS1 client and configure it for the mock service
+	ns1 := api.NewClient(doer, api.SetAPIKey("apikey"))
+	ns1.Endpoint, _ = url.Parse("https://" + mock.Address + "/v1/")
+
+	// Add your test case (zone list in this example)
+	require.Nil(t, mock.AddTestCase(http.MethodGet, "zones", http.StatusOK, nil, nil, "",
+		[]*dns.Zone{{Zone: "foo.bar"}}))
+
+	// Perform your tests
+	zones, _, err := ns1.Zones.List()
+	require.Nil(t, err)
+	require.Equal(t, 1, len(zones))
+	require.Equal(t, "foo.bar", zones[0].Zone)
+
+	// Output:
+}
--- a/mockns1/servehttp.go
+++ b/mockns1/servehttp.go
@@ -1,0 +1,111 @@
+package mockns1
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// ServeHTTP is the request  handler for the mock service. This
+// should not be called directly in unit tests.
+func (s *Service) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	s.stopTimer()
+	defer s.startTimer()
+
+	if _, exists := s.tests[r.Method]; !exists {
+		notFoundResponse(w, "method")
+		return
+	}
+
+	tests, exists := s.tests[r.Method][r.RequestURI]
+	if !exists {
+		notFoundResponse(w, "uri")
+		return
+	}
+
+	defer r.Body.Close()
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		w.Write([]byte( // nolint: errcheck
+			fmt.Sprintf(`{"message": "unable to read request body: %s`, err),
+		))
+		return
+	}
+
+	var test *testCase
+	for _, t := range tests {
+		if !assert.Equal(new(testifyT), t.request.body, body) {
+			continue
+		}
+
+		if !compareHeaders(t.request.headers, r.Header) {
+			continue
+		}
+
+		test = t
+		break
+	}
+
+	if test == nil {
+		notFoundResponse(w, "no test")
+		return
+	}
+
+	for k, vals := range test.response.headers {
+		w.Header().Set(k, vals[0])
+		for _, v := range vals[1:] {
+			w.Header().Add(k, v)
+		}
+	}
+
+	w.WriteHeader(test.status)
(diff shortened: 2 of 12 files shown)


mburtless:
👋 Hi @iamgnat - This looks great, thanks for the contribution!  Definitely something that folks have requested in the past and should streamline some of the tests in this repo.  I'll find some time this week to take a look and leave a review.

mockns1/example_test.go
```diff
@@ -0,0 +1,38 @@
...
+	require.Nil(t, mock.AddTestCase(http.MethodGet, "zones", http.StatusOK, nil, nil, "",
+		[]*dns.Zone{{Zone: "foo.bar"}}))
+
+	// Perform your tests
+	zones, _, err := ns1.Zones.List()
+	require.Nil(t, err)
+	require.Equal(t, 1, len(zones))
+	require.Equal(t, "foo.bar", zones[0].Zone)
+
+	// Output:
```
mburtless, line 37: Is this comment still needed?

mockns1/testcase.go
```diff
@@ -0,0 +1,92 @@
...
+		s.tests[method] = map[string][]*testCase{}
+	}
+	if _, exists := s.tests[method][uri]; !exists {
+		s.tests[method][uri] = []*testCase{}
+	}
+
+	t := new(testifyT)
+	for _, test := range s.tests[method][uri] {
+		header := assert.Equal(t, tc.request.headers, test.request.headers)
+		body := assert.Equal(t, tc.request.body, test.request.body)
```
mburtless, line 65: cool use of assert.Equal, it seems so obvious but I've never used it this way before

mockns1/servehttp.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+
+		test = t
+		break
+	}
+
+	if test == nil {
+		notFoundResponse(w, "no test")
+		return
+	}
```
mburtless: Maybe just a personal thing, but `continue`s and nested `if`s always take me a minute to comprehend.  Maybe this would be easier to read as a seperate function. ie
```
func matchTest(tests map[string]map[string][]*testCase) error {
	for _, t := range tests {
		if (t.request.json && assert.JSONEq(new(testifyT), string(t.request.body), string(body))) ||
		(!t.request.json && assert.Equal(new(testifyT), t.request.body, body))  {
			if compareHeaders(t.request.headers, r.Header) {
				return t
			}
		}
	}

	return notFoundResponse(w, "no test")
}
```

mburtless (dismissed):
Couple of minor comments, but this all looks great to me.  None are critical imo, so I'll merge this as is on Monday, if you don't have time to address them or don't think they are necessary.  Thanks again for taking the time to contribute this work.

Just as a heads up, as a company we're working internally to utilize OpenAPI spec across much more of our platform, including SDK and mock generation.  Due to some competing priorities, we probably won't be able to overhaul this SDK with a completely generated version in the near future, but some lower hanging fruit could include using this work in combination with our spec to generate mocks.  Unfortunately, we probably can't commit to a timeline for publishing our spec and beginning this work, but will try to post updates to this repo as that progresses.

Given the prioritization above, we probably won't be able to devote much time immediately to manually fleshing out this mock, but of course we're more than happy to review any contributions to this in the mean time.

mockns1/example_test.go
iamgnat: That is a function of Go's example functionality. In addition to generating the example for the documentation, it uses the `// Output:` comment to determine if the example was successful or not (it executes from `go test` along with the other unit tests).

In this case the example program does not write any output so the comment does look a little funny. If it was writing output it would look something like:

```go
fmt.Println("Hello world!")

// Output:
// Hello world!
```

[Go's explanation](https://blog.golang.org/examples)

mburtless: ah, makes sense. TIL!

mockns1/servehttp.go
iamgnat: Heh. This is what I get for trying to re-write code from memory rather than looking at my original code (Salesforce.com internal version that I've gotten permission to port to this project)...

I'm not a fan of long `if` statements like your example as I find them less readable, but I can definitely clean up `ServeHTTP()` by breaking things up better like my original code.

Please hold off on the merge, but I'll try to get this taken care of this weekend anyway. Thanks for pointing this out.

iamgnat: @mburtless Here is my take on making that loop read easier. Have a look when you have a chance.

mburtless: Cool solution, that looks great.

mburtless (approved):
Thanks again for the contribution, can't wait to use this!

PR #105, opened 2020-06-30
bump version to fix tagging issue

Zach-Johnson:
Need to release 2.4.1 to re-tag it, I think 2.4.0 was pointed to an older version of master


mburtless (approved):
🙇

PR #106, opened 2020-07-30
support for named zones/dns views

rupa:
from a views POV:
  * where thing take a "zone" argument, this is used as the "name".
    I haven't renamed the variables, but maybe i should?
  * existing zone and record create funcs set both the FQDN and name
    to the passed "zone"
  * new zone and record create funcs added for passing both handle and name.
    you could also edit the object returned by the "old" function.
  * crud funcs that take a model use GetName/GetZoneName when building
    url
  * As noted, other funcs should all work as normal, with the caveat
    that if "name/zoneName" != FQDN, the name must be passed..

zone model gets:
  * Name attr
  * Views list attr
  * String method shows name, if present
  * GetName method (rather than try to keep the object sync'd)

record model gets:
  * Name attr
  * String method shows zoneName, if present
  * GetZoneName method (rather than try to keep the object sync'd)

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,87 @@ func main() {
 }
 ```
 
+DNS views and compatibility
+===========================
+
+DNS views is a means for NS1 to serve one set of data to one group of clients
+(e.g. internal employees), and different sets of data to other groups of
+clients (e.g. public internet). This has been largely exposed by allowing zones
+in the NS1 system to share the same FQDN, and allowing propagation to be
+controlled via ACLs and "views". For existing zones, and users that have no
+need for the added complexity of views, the default behavior is unchanged.
+However, it is important to understand that the requirement that an FQDN be
+unique within the network is removed in v3.x, and the ramifications of that.
+
+Views are available in API v3.x. For more information, see:
+https://help.ns1.com/hc/en-us/articles/360054071374
+
+COMPATIBILITY
+
+Since an FQDN can now appear in more than one "zone", it can no longer uniquely
+identify a zone. Instead, a user-supplied "name", unique within an account,
+is used to uniquely identify the zone.
+
+The zone name can be the same as its FQDN, existing zones transfer to the new
+schema this way. And if not using views, it does serve as a good identifier.
+However, if a second zone is created pointing to same FQDN, it cannot reuse
+the FQDN as an identifier, and queries for the zone FQDN (as an identfier)
+will return the first zone. The following example should help illustrate:
+
+In general, for API requests, the identifier for a zone will appear in the URL.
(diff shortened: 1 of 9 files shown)


rest/model/dns/zone.go
```diff
@@ -14,7 +14,10 @@ type Zone struct {
 	Pool         string   `json:"pool,omitempty"` // Deprecated
 
 	ID   string `json:"id,omitempty"`
-	Zone string `json:"zone,omitempty"`
+	Zone string `json:"zone,omitempty"` // FQDN
+	// Unique identifier for the zone within account. In most cases, it is best
+	// to also use the zone FQDN for this.
+	Name string `json:"name,omitempty"`
```
Zach-Johnson, line 20: just my 2 cents, I think I'd rather this be `ZoneName` to stay consistent with `Record`

rupa: I'd agree, but this matches the API, where it's name on zone, and zone_name on records

README.md
```diff
@@ -53,6 +53,87 @@ func main() {
...
+
+However, in 3.x we are now allowed to make new zones with the same FQDN:
+
+$ENDPOINT/v1/zones/example-internal -d '{zone: example.com, name: example-internal}'
+                   ^                     ^                  ^
+                   L___ NAME             L____ ZONE         L____ NAME
+
+`example-internal` shares the FQDN with `example.com`. API calls using
+`example.com` as the identifier will uniquely identify the first zone, to
+address the second zone, `example-internal` must be used in the identifier
```
jorge-gautier, line 115: It would read better to put a period or a semicolon after `the first zone`.

README.md
```diff
@@ -53,6 +53,87 @@ func main() {
...
+How they are propagated relies on how they are organized with regard to views,
+acls, and networks.
+
+SUMMARY OF CHANGES TO THE SDK FOR VIEWS
+
+For compatibility, we've left the `zone` variable alone in existing functions.
+When name and FQDN differ, and you are calling a func that takes `zone`, you
+want to pass the identifier, `name`. The only time you need to be sure to pass
+the zone's FQDN is on zone or record creation, where we need to know it.
+Otherwise, we're doing lookups, and the `name` is the unambigious field.
```
jorge-gautier: Typo, should be `unambiguous`.

jorge-gautier:
Code LGTM. Small issues in the `README.md`.

README.md
rupa: done

rupa: done

jorge-gautier (approved):

PR #111, opened 2020-08-14
update config for https jobs

Haugum:


rest/model/monitor/job.go
```diff
@@ -121,13 +123,18 @@ type Rule struct {
 // ua is the user agent text in the request header.
 // auth is the authorization header to use in request.
 // connTimeout is the timeout(in sec) to wait for query output.
-func NewHTTPConfig(url, method, ua, auth string, connTimeout int) *Config {
+func NewHTTPConfig(url, method, ua, auth string, connTimeout int, it time.Duration, reqIPV4 bool, vhost string, tlsSkipVerify bool, followRedir bool) *Config {
```
Zach-Johnson: @Haugum what's the context for this change? We're not going to be able to merge this into v2 as this is a breaking change (changing a function signature on a public method), so we either need to re-think how we're doing this or table it for a major release.

cpg1111 (approved):
I'd also get @Zach-Johnson to take a peak, but LGTM

Zach-Johnson (approved):
sweet!

PR #116, opened 2020-11-25
DHCP-312/add dhcp/optiondef methods

elenabushneva:


rest/model/dhcp/option.go
```diff
@@ -0,0 +1,85 @@
...
+	Fields             []OptionDefSchemaItems `json:"fields"`
+	Items              *string                `json:"items,omitempty"`
+	MultipleFinalValue *bool                  `json:"multiple_final_value,omitempty"`
+	Type               SchemaType             `json:"type"`
+}
+
+// OptionDef configures a custom option definition
+// https://ftp.isc.org/isc/kea/1.4.0/doc/kea-guide.html#dhcp4-custom-options
+type OptionDef struct {
+	Space        string          `json:"space,omitempty"`
```
cbertinato: Could this ever be empty?

cbertinato (dismissed):
This looks great. There's just that one question. If there is no change incoming as a result, then this is good to go.

cbertinato (approved):

PR #119, opened 2020-12-29
DNS tagging support

dwhittle-sfdc:
Adding `Tags`, `BlockedTags`, and `LocalTags` fields to `Zone` and `Record` DNS objects to support new functionality in DDI 3.2.3

### DIFF
--- a/rest/model/dns/record.go
+++ b/rest/model/dns/record.go
@@ -27,6 +27,16 @@ type Record struct {
 	Filters []*filter.Filter `json:"filters"`
 	// The records' regions.
 	Regions data.Regions `json:"regions,omitempty"`
+
+	// Contains the key/value tag information associated to the record
+	Tags map[string]string `json:"tags,omitempty"`
+
+	// List of tag key names that should not inherit from the parent zone
+	BlockedTags []string `json:"blocked_tags,omitempty"`
+
+	// List of tag key namess set directly on this record. Any other Tag entries
+	// are inherited from the parent zone
+	LocalTags []string `json:"local_tags,omitempty"`
 }
 
 func (r Record) String() string {
@@ -40,13 +50,16 @@ func NewRecord(zone string, domain string, t string) *Record {
 		domain = fmt.Sprintf("%s.%s", domain, zone)
 	}
 	return &Record{
-		Meta:    &data.Meta{},
-		Zone:    zone,
-		Domain:  domain,
-		Type:    t,
-		Answers: []*Answer{},
-		Filters: []*filter.Filter{},
-		Regions: data.Regions{},
+		Meta:        &data.Meta{},
+		Zone:        zone,
+		Domain:      domain,
+		Type:        t,
+		Answers:     []*Answer{},
+		Filters:     []*filter.Filter{},
+		Regions:     data.Regions{},
+		Tags:        map[string]string{},
+		BlockedTags: []string{},
+		LocalTags:   []string{},
 	}
 }
 
--- a/rest/model/dns/record.go
+++ b/rest/model/dns/record.go
@@ -29,14 +29,13 @@ type Record struct {
 	Regions data.Regions `json:"regions,omitempty"`
 
 	// Contains the key/value tag information associated to the record
-	Tags map[string]string `json:"tags,omitempty"`
+	Tags map[string]string `json:"tags,omitempty"` // Only relevant for DDI
 
 	// List of tag key names that should not inherit from the parent zone
-	BlockedTags []string `json:"blocked_tags,omitempty"`
+	BlockedTags []string `json:"blocked_tags,omitempty"` //Only relevant for DDI
 
-	// List of tag key namess set directly on this record. Any other Tag entries
-	// are inherited from the parent zone
-	LocalTags []string `json:"local_tags,omitempty"`
+	// Read-only fields
+	LocalTags []string `json:"local_tags,omitempty"` // Only relevant for DDI
 }
 
 func (r Record) String() string {
@@ -59,7 +58,6 @@ func NewRecord(zone string, domain string, t string) *Record {
 		Regions:     data.Regions{},
 		Tags:        map[string]string{},
 		BlockedTags: []string{},
-		LocalTags:   []string{},
 	}
 }
 
--- a/rest/model/dns/record.go
+++ b/rest/model/dns/record.go
@@ -49,15 +49,13 @@ func NewRecord(zone string, domain string, t string) *Record {
 		domain = fmt.Sprintf("%s.%s", domain, zone)
 	}
 	return &Record{
-		Meta:        &data.Meta{},
-		Zone:        zone,
-		Domain:      domain,
-		Type:        t,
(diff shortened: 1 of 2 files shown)


rest/model/dns/zone.go
```diff
@@ -110,7 +133,10 @@ type TSIG struct {
 // NewZone takes a zone domain name and creates a new zone.
 func NewZone(zone string) *Zone {
 	z := Zone{
-		Zone: zone,
+		Zone:        zone,
+		Tags:        map[string]string{},
+		BlockedTags: []string{},
+		LocalTags:   []string{},
```
mburtless: This will only work for creating a new zone in DDI, but this SDK supports both DDI and managed DNS.

Is this required for the constructor in DDI?  The API will handle empty tags/blocked_tags fields on POST.  `local_tags` is read only and should not be included.

rest/model/dns/zone.go
```diff
@@ -41,6 +44,16 @@ type Zone struct {
...
+
+	// Contains the key/value tag information associated to the zone
+	Tags map[string]string `json:"tags,omitempty"`
+
+	// List of tag key names that should not inherit from the parent object
+	BlockedTags []string `json:"blocked_tags,omitempty"`
+
+	// List of tag key namess set directly on this zone. Any other Tag entries
+	// are inherited from a higher level
+	LocalTags []string `json:"local_tags,omitempty"`
```
mburtless: This field should probably live in section of struct with `Read-only fields` comment, above this line.

rest/model/dns/zone.go
```diff
@@ -41,6 +44,16 @@ type Zone struct {
 	// Whether or not DNSSEC is enabled on the zone. Note we use a pointer so
 	// leaving this unset will not change a previous setting.
 	DNSSEC *bool `json:"dnssec,omitempty"`
+
+	// Contains the key/value tag information associated to the zone
+	Tags map[string]string `json:"tags,omitempty"`
+
+	// List of tag key names that should not inherit from the parent object
```
mburtless: Comments should note that these fields are only relevant in DDI and should not be set for managed, as with [other DDI only fields](https://github.com/ns1/ns1-go/blob/700d54285d1a69bdd4b0e0e72acd8bf2fe36d2c4/rest/model/account/permissions.go#L11)

rest/model/dns/record.go
```diff
@@ -40,13 +50,16 @@ func NewRecord(zone string, domain string, t string) *Record {
...
+		Meta:        &data.Meta{},
+		Zone:        zone,
+		Domain:      domain,
+		Type:        t,
+		Answers:     []*Answer{},
+		Filters:     []*filter.Filter{},
+		Regions:     data.Regions{},
+		Tags:        map[string]string{},
+		BlockedTags: []string{},
+		LocalTags:   []string{},
```
mburtless: This will only work for creating a new record in DDI, but this SDK supports both DDI and managed DNS.

Is this required for the constructor in DDI? The API will handle empty tags/blocked_tags fields on POST. local_tags is read only and should not be included.

rest/model/dns/zone.go
```diff
@@ -41,6 +44,16 @@ type Zone struct {
...
+	BlockedTags []string `json:"blocked_tags,omitempty"`
```
mburtless: Blocked tags are not support at the zone level

rest/model/dns/zone.go
```diff
@@ -56,6 +69,16 @@ type ZoneRecord struct {
...
+
+	// Contains the key/value tag information associated to the record
+	Tags map[string]string `json:"tags,omitempty"`
+
+	// List of tag key names that should not inherit from the parent zone
+	BlockedTags []string `json:"blocked_tags,omitempty"`
+
+	// List of tag key namess set directly on this record. Any other Tag entries
+	// are inherited from the zone
+	LocalTags []string `json:"local_tags,omitempty"`
```
mburtless: Same deal as with zone tags, `LocalTags` should move to the read only section and the comments on all tags fields should denote they are only applicable to DDI.

mburtless (changes requested):
Hi @dwhittle-sdfc - Thanks so much for the contribution!  Apologies for the delayed response here, most of our reviewers for this repo were out of pocket due to the holidays.

This is looking good, but I think this needs a couple of small changes before we can merge it.  The main thing to keep in mind here is that this SDK is used for both DDI and managed DNS, but not all features exist on both platforms (ie tagging).  This means we need to implement shared funcs like constructors in such a way that they continue to work for both platforms and fields in shared structs that are specific to a single platform should be clearly denoted via comments.

dwhittle-sfdc:
@mburtless Thanks for the review. I have submitted updates based on your comments. Let me know if more is needed.

rest/model/dns/zone.go
```diff
@@ -111,6 +128,7 @@ type TSIG struct {
 func NewZone(zone string) *Zone {
 	z := Zone{
 		Zone: zone,
+		Tags: map[string]string{},
```
mburtless: `tags` is not a valid field for a new zone in managed DNS.  Can this be removed from the constructor?

mburtless (changes requested):
Looking good!  I think we should still remove the `tags` field from the zone returned by `NewZone` constructor though, as that will not be compatible with managed.

rest/model/dns/zone.go
dwhittle-sfdc: From a user's perspective it would be expected that after calling a constructor that the entire structure is initialized appropriately. 

If `Tags` is not initialized here the user then tries to add something to it, a `panic` will be generated: https://play.golang.com/p/vVtxeKFaUaZ

I believe this is more idiomatic and user friendly, but I do understand your concerns given the dual functionality that this module needs to support. I'll remove it if you still want me to.

mburtless: I'm with you on this being more user friendly, but this effectively breaks existing functionality as every request that's submitted to managed using a zone initialized by this constructor will contain an invalid field after this PR is merged.  Our API silently ignores extra fields, but this should still be avoided.

How about removing the field from this constructor, and instead adding a new, DDI specific, constructor (ie `NewDDIZone`) that includes initialization of the tags map?

dwhittle-sfdc: I'll just remove it for now and leave a `NewDDIZone()` construction for another time as similar constructors should be probably be done for other objects (Record, Permissions, etc..) that have similar variances. 

Another option might be to add helpers around the tags (`Set`, `Get`, `IsLocal`, etc..) which could hide the need to consider if the attributes are already initialized.

dwhittle-sfdc: Removed the `Tags` and `BlockedTags` from both `Record` and `Zone` constructors.

mburtless: LGTM.  Love the tag helper idea, could even be generalized to work on all resources with tagging functionality (ie ipam objects as well).

mburtless (approved):
LGTM, thanks again for the contribution.  Will cut a release for this today.

PR #123, opened 2021-03-31
Enable support for DNSSec record type DS

juli-vert:


rest/model/dns/answer.go
```diff
@@ -141,6 +141,19 @@ func NewSRVAnswer(priority, weight, port int, target string) *Answer {
 	}
 }
 
+// NewDSAnswer creates an Answer for DS record.
+func NewDSAnswer(key string, algorithm string, t string, digest string) *Answer {
+	return &dns.Answer{
```
lmower: Looks like the build is failing https://travis-ci.org/github/ns1/ns1-go/builds/765312079 - I think you want `&dns.Answer`  to be `&Answer`

juli-vert: Sure! My fault! copy/pasting my local solution was not a good idea. Commit pushed

ponbiki (approved):
lgtm

PR #125, opened 2021-04-01
Add auth tag support

mburtless:
This PR adds support for IPAM/DHCP `tags_allow` and `tags_deny` permissions, which are available in DDI versions 2.5.x and above.

The fields have been added as pointers to an array to satisfy the following two requirements:
1. Empty list is a valid value (the default value for these fields)
2. The fields may only included in requests when explicitly set, to ensure backwards compatibility with DDI versions below 2.5.x that do not support this field

This PR also adds a new example which shows how this SDK, as well as these new permissions, can be used with DDI.


rest/model/account/permissions.go
```diff
@@ -62,10 +62,23 @@ type PermissionsMonitoring struct {
...
+	TagsAllow *[]AuthTag `json:"tags_allow,omitempty"`
+	TagsDeny  *[]AuthTag `json:"tags_deny,omitempty"`
 }
 
 // PermissionsIPAM wraps a User's "permissions.ipam" attribute for DDI.
 type PermissionsIPAM struct {
 	ManageIPAM bool `json:"manage_ipam"`
 	ViewIPAM   bool `json:"view_ipam"`
+	// The fields below are only relevant in DDI v2.5+
+	TagsAllow *[]AuthTag `json:"tags_allow,omitempty"`
```
lmower, line 75: A bit of a nit, but I don't _think_ we need a pointers to slices as an empty slice can evaluate as equal to `nil` which will satisfy the `omitempty` decorator.

rest/account_apikey_test.go
```diff
@@ -68,4 +88,43 @@ func TestCreateDDIAPIKey(t *testing.T) {
 
 	_, err := c.APIKeys.Create(k)
 	require.NoError(t, err)
+	// Create a key with auth tags
+	k = &account.APIKey{
```
lmower, line 101: Might also be valuable to include a test with a key where permissions are empty like:
```go
	account.APIKey{
		ID:   "foo",
		Key:  "foo-1",
		Name: "foo-1",
		Permissions: account.PermissionsMap{
			DHCP: &account.PermissionsDHCP{},
			IPAM: &account.PermissionsIPAM{},
		},
	}
```

lmower:
Could you run `go fmt`? There appears to be some minor whitespace changes in `rest/account_apikey_test.go`

rest/model/account/permissions.go
mburtless: That's the problem we are trying to solve with the pointer.
* with `omitempty` and not using a pointer an empty slice will always be omited.  This is not what we want as empty slice is a valid value for this field and must be able to be included in a request or recieved in a response.
* without `omitempty` and not using a pointer. `TagsAllow` will _always_ be included in request, when the user does not explicitly define it the zero value will be used (empty slice).  This is also not what we want as this field is not valid in DDI versions below v2.5.

rest/account_apikey_test.go
mburtless: good call!

rest/model/account/permissions.go
lmower: I see, thanks for explaining.

lmower (approved):
🏆

PR #127, opened 2021-05-20
PERSONA-180 run rate limit when there is a response

decarvalhorobinson:
When an error occurs the client will not run the rate limit, which causes too many requests to be fired and then will hit the endpoints rate limit. So to avoid this, the rate limit function will be run when there is a response from the client payload. This was noticed when running terraform provider with invalid fields, which caused 400 errors

### DIFF
--- a/rest/client.go
+++ b/rest/client.go
@@ -164,16 +164,13 @@ func SetDDIAPI() func(*Client) {
 // non-2XX response.
 func (c Client) Do(req *http.Request, v interface{}) (*http.Response, error) {
 	resp, err := c.httpClient.Do(req)
-
-	if resp != nil {
-		rl := parseRate(resp)
-		c.RateLimitFunc(rl)
-	}
-
 	if err != nil {
 		return nil, err
 	}
 	defer resp.Body.Close()
+
+	rl := parseRate(resp)
+	c.RateLimitFunc(rl)
 
 	err = CheckResponse(resp)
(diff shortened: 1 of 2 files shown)


mburtless (approved):
Such a simple solution, I love this.

rest/client.go
```diff
@@ -164,6 +164,12 @@ func SetDDIAPI() func(*Client) {
 // non-2XX response.
 func (c Client) Do(req *http.Request, v interface{}) (*http.Response, error) {
 	resp, err := c.httpClient.Do(req)
+
+	if resp != nil {
```
mburtless: Can we move this check below the `if err != nil` beneath this? I don't think there's a reason to eschew Go's convention for error checking immediately after the func call, since HTTP error status code does not result in `httpClient.Do` returning an `err`, right?

mburtless (changes requested):
Actually, one question on this

rest/client.go
decarvalhorobinson: If the "doer" doesn't return any errors for the requests that have responses, it makes sense to move it down after the first error checking. I am testing, if everything works well I will make another commit moving it

decarvalhorobinson: I moved since it seems what you said is correct, so it makes senses to keep it after the first error checking

mburtless (approved):
:fire:

PR #129, opened 2021-06-25
MON-117 add subdivisions support

pcatalini:


Haugum (dismissed):

mburtless (changes requested):
This change makes sense, but we should add unit tests to ensure it is working as expected and we don't regress.  There are existing tests for `MetaFromMap` and `StringMap` that can be extended here.

mburtless (approved):
:fire:

PR #137, opened 2021-07-16
More verbose error output when encoutering a JSON decoding error.

AskAlice:
Getting more information about what was actually sent to the client would be vital for understanding the fault itself in the event of a JSON decoding error.

This returns output explaining the response headers and response body of what was actually returned given an API call.

### DIFF
--- a/rest/client.go
+++ b/rest/client.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


AskAlice:
We initially tried to do this on our end, but since the request body was already closed in the ns1-go client, it is not possible to do this on the end-user side.

rest/client.go
```diff
@@ -179,8 +180,12 @@ func (c Client) Do(req *http.Request, v interface{}) (*http.Response, error) {
 
 	if v != nil {
 		// Try to unmarshal body into given type using streaming decoder.
-		if err := json.NewDecoder(resp.Body).Decode(&v); err != nil {
-			return nil, err
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return nil, fmt.Errorf("error reading response body: %w", err)
+		}
```
mburtless, line 186: I think we already perform this in `CheckResponse`, which seems like the more appropriate place to check for json decoding errors.  Can we make the error returned from that method more verbose instead of making this section more complex?

mburtless (changes requested):
:wave: Hi!  Thanks so much for the contribution, always good to be more explicit about these types of errors.

I'm on board with the change, but I think the more appropriate location might be the `CheckResponse` func where we already perform these checks. What do you think?

rest/client.go
pcatalini: @AskAlice How about following @mburtless suggestion, does it make sense?

eravin-ns1:
This has been fixed in PR #177 and is now available in release 2.7.2.

PR #148, opened 2021-10-26
DAITDS-247 - Added defaultPermissions function

ThiagoYU:


rest/account_apikey_v2.go
```diff
@@ -0,0 +1,184 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var a account.APIKeyV2
+	resp, err := s.client.Do(req, &a)
+	if err != nil {
+		switch err.(type) {
+		case *Error:
+			if err.(*Error).Message == "unknown api key" {
```
Zach-Johnson:
``` suggestion
		switch e := err.(type) {
		case *Error:
			if e.Message == "unknown api key" {
```

rest/account_apikey_v2.go
```diff
@@ -0,0 +1,184 @@
...
+	return resp, nil
+}
+
+// Same as v1
+// var (
+// 	// ErrKeyExists bundles PUT create error.
+// 	ErrKeyExists = errors.New("key already exists")
+// 	// ErrKeyMissing bundles GET/POST/DELETE error.
+// 	ErrKeyMissing = errors.New("key does not exist")
+// )
```
Zach-Johnson: lets just delete this stuff then

rest/model/account/permissions_v2.go
```diff
@@ -0,0 +1,94 @@
...
+// 	Name  string `json:"name"`
+// 	Value string `json:"value"`
+// }
+
+// type PermissionsRecord struct {
+// 	Domain     string `json:"domain"`
+// 	Subdomains bool   `json:"include_subdomains"`
+// 	Zone       string `json:"zone"`
+// 	RecordType string `json:"type"`
+// }
```
Zach-Johnson: lets delete this

Zach-Johnson (changes requested):
@ThiagoYU the permissions piece of this looks good, I had just a couple of small comments.

On the duplication for the user/team/apikey objects, I'm wondering if we can do something different to avoid this much duplication. Do you think we could put the PermissionsMapV2 on the v1 structs for user/team/apikey as a pointer and then use a custom json marshaler to choose which permission struct to use? So basically, if the PermissionsMapV2 != nil, then use that, otherwise use the original. I _think_ this could work, and would remove a lot of the duplication. What do you think?

ThiagoYU:
This seems to be fixed

PR #165, opened 2022-09-29
INBOX-2163 fix zone already exists error

achunariov:


eravin-ns1 (dismissed):
LGTM

Zach-Johnson (dismissed):
LGTM, I'd rather us update this to check a status code from the API response to handle this, but I think we don't return the right status code from the API for this case right now.

eravin-ns1 (approved):

PR #147, opened 2021-10-19
DAITDS-240 - DNS View added

ThiagoYU:

PR #171, opened 2022-10-19
Dns views

eravin-ns1:
* adds support for DNS views - this is an updated version of #147 but now up to date with current v2 branch


README.md
```diff
@@ -53,6 +53,87 @@ func main() {
...
+===========================
+
+DNS views is a means for NS1 to serve one set of data to one group of clients
+(e.g. internal employees), and different sets of data to other groups of
+clients (e.g. public internet). This has been largely exposed by allowing zones
+in the NS1 system to share the same FQDN, and allowing propagation to be
+controlled via ACLs and "views". For existing zones, and users that have no
+need for the added complexity of views, the default behavior is unchanged.
+However, it is important to understand that the requirement that an FQDN be
+unique within the network is removed in v3.x, and the ramifications of that.
```
Zach-Johnson: this sounds like a ddi specific comment, I think it should be removed

README.md
```diff
@@ -53,6 +53,87 @@ func main() {
...
+
+Views are available in API v3.x. For more information, see:
```
Zach-Johnson: same

README.md
```diff
@@ -53,6 +53,87 @@ func main() {
...
+https://help.ns1.com/hc/en-us/articles/360054071374
+
+COMPATIBILITY
```
Zach-Johnson: this seems like generally more information than needs to be in the SDK, this should be captured in API docs if it isn't. I don't think all of this belongs here. If you disagree, then all of the ddi-specific language needs to be changed.

Zach-Johnson (changes requested):

README.md
eravin-ns1: I think we need a short discussion of views and how view-aware code differs from older code, and the rest can be dropped.

eravin-ns1: README is updated, PTAL

rest/client.go
```diff
@@ -12,7 +12,7 @@ import (
 )
 
 const (
-	clientVersion = "2.6.5"
+	clientVersion = "2.6.6"
```
Zach-Johnson: changelog should be updated alongside a version bump. I would suggest keeping these in a separate PR rather than in a feature PR

eravin-ns1: Done.

Zach-Johnson (approved):

Issue #117, opened 2020-12-10
License discrepancy with https://pkg.go.dev/


SelfDrivingCarp:
https://pkg.go.dev/ is the new system for browsing package documentation.

https://pkg.go.dev/gopkg.in/ns1/ns1-go.v2 to read this package because of a discrepancy with the license. The license file is there in a file name the site expects so there must be a discrepancy with the license text.

The license policy is here: https://pkg.go.dev/license-policy

pcatalini:
@jasonmf  Thank you for reporting this issue.
We already added this to our backlog.

eravin-ns1:
The doc page at https://pkg.go.dev/gopkg.in/ns1/ns1-go.v2 now works properly.

PR #172, opened 2022-10-19
INBOX-2191 / pass the golang license check

eravin-ns1:
The Golang automation uses the [licensecheck](https://pkg.go.dev/github.com/google/licensecheck) package to detect the type of license. This PR removes text after the "END OF TERMS AND CONDITIONS" line that the license checker did not expect to see. 

The problem is visible at https://pkg.go.dev/gopkg.in/ns1/ns1-go.v2  which refuses to display this package's documentation due to the license check failure.


cbertinato (approved):

eravin-ns1:
Resolves #117

PR #174, opened 2022-12-05
don't trust received rate-limit headers

eravin-ns1:
If the rate-limit header values in the API response are missing or corrupt, provide a default value of 100 ms. This will lead to a sleep of 100 ms times the "parallelism" user-settable value. For more details see the [NS1 Terraform provider documentation](https://registry.terraform.io/providers/ns1-terraform/ns1/latest/docs#rate_limit_parallelism).


ponbiki (approved):
Seems fine to me (at least until we find the source of the 0 limit and/or period header values)

PR #179, opened 2022-12-13
release 2.7.2

eravin-ns1:
## 2.7.2 (December 13, 2022)
BUG FIXES:

* When an HTTP error response with a non-JSON body is received, use the body as the error message (instead of a message about being unable to parse JSON).


bricaurte-NS1 (approved):
+1

PR #195, opened 2023-03-03
feat: allow empty tags in record/zone dns model

fgouteroux:
Hello,

In order to add the support of tags in `ns1_zone` and `ns1_record` of the [terraform-provider-ns1](https://github.com/ns1-terraform/terraform-provider-ns1), we need to allow setting empty tags in the record and zone dns model. See https://github.com/ns1-terraform/terraform-provider-ns1/pull/279

Could you have a review please ?

### DIFF
--- a/rest/model/dns/record.go
+++ b/rest/model/dns/record.go
@@ -30,10 +30,10 @@ type Record struct {
 	Regions data.Regions `json:"regions,omitempty"`
 
 	// Contains the key/value tag information associated to the record
-	Tags map[string]string `json:"tags,omitempty"` // Only relevant for DDI
+	Tags map[string]string `json:"tags"` // Only relevant for DDI
 
 	// List of tag key names that should not inherit from the parent zone
-	BlockedTags []string `json:"blocked_tags,omitempty"` //Only relevant for DDI
+	BlockedTags []string `json:"blocked_tags"` //Only relevant for DDI
 
 	// Read-only fields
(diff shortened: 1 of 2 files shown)


fgouteroux:
Hello, any news about this ?

ponbiki (dismissed):

rest/model/dns/record.go
```diff
@@ -30,10 +30,10 @@ type Record struct {
 	Regions data.Regions `json:"regions,omitempty"`
 
 	// Contains the key/value tag information associated to the record
-	Tags map[string]string `json:"tags,omitempty"` // Only relevant for DDI
```
ponbiki, line 33: These tags are actually omitted on the zones CRUD routes, unless explicitly instantiated.

ponbiki (commented):
The tags in the model are omitempty, as certain routes do intentionally omit them at the API level unless the user explicitly wants to instantiate them.

fgouteroux:
@ponbiki The ns1 api don't remove zone/records tags if the json key tags is not present. As [terraform-provider-ns1](https://github.com/ns1-terraform/terraform-provider-ns1) is using this library and the dns.Zone/dns.Record model, so we have no control on the marshall/unmarshall json object.

I understand this is a breaking change and maybe it's not the right way. In my opinion, when the api received the payload without json tags key, it should remove all record/zone tags.

PR #214, opened 2023-07-27
Remove ID field from monitor job create path

ponbiki:
This PR removes the ID field from the URL for monitor creation.


hmodi-ns1 (dismissed):
Some tests should be good, Approving as this is required for a hotfix.

hmodi-ns1 (approved):
lgtm

PR #222, opened 2023-10-12
v2.7.13

ponbiki:
Removes `omitempty` from `tags`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 2.7.13 (October 12th, 2023)
+
+BUG FIXES:
+
+* Remove `omitempty` from `tags` to allow POST without `tags` after update 2.7.12
+
 ## 2.7.12 (October 11th, 2023)
 
 BUG FIXES:
--- a/rest/client.go
+++ b/rest/client.go
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	clientVersion = "2.7.12"
+	clientVersion = "2.7.13"
 
 	defaultEndpoint               = "https://api.nsone.net/v1/"
 	defaultShouldFollowPagination = true
--- a/rest/model/dns/record.go
+++ b/rest/model/dns/record.go
@@ -31,7 +31,7 @@ type Record struct {
 	Regions data.Regions `json:"regions,omitempty"`
 
 	// Contains the key/value tag information associated to the record
-	Tags map[string]string `json:"tags,omitempty"` // Only relevant for DDI
+	Tags map[string]string `json:"tags"` // Only relevant for DDI
 
 	// List of tag key names that should not inherit from the parent zone
 	BlockedTags []string `json:"blocked_tags"` //Only relevant for DDI


tlkamp:
@ponbiki hello, the removal of `omitempty` introduced a breaking change to `terraform-provider-ns1` when it was removed from blocked tags. 

Find the attached issue comment: https://github.com/ns1-terraform/terraform-provider-ns1/issues/304#issuecomment-1760666592

I think this PR will introduce the same issue, but now for tags.

ponbiki:
> @ponbiki hello, the removal of `omitempty` introduced a breaking change to `terraform-provider-ns1` when it was removed from blocked tags.
> 
> Find the attached issue comment: [ns1-terraform/terraform-provider-ns1#304 (comment)](https://github.com/ns1-terraform/terraform-provider-ns1/issues/304#issuecomment-1760666592)
> 
> I think this PR will introduce the same issue, but now for tags.

The omitempty needed removed from both unfor

tlkamp:
> The omitempty needed removed from both unfor

@ponbiki Looks like your message was incomplete. Why do they both need removed? Are `tags` and `blocked_tags` now required fields of the API? The documentation does not seem to indicate that.

https://ns1.com/api?docId=2185

The removal of the `omitempty` struct tag is causing your client to send the `blocked_tags` field to the API even when it has no content now where it was not sent before, resulting in the error I reported:

```
Error: POST https://api.nsone.net/v1/zones/MYZONE/MYRECORD/A: 400 tags and blocked_tags must be updated together
```

I believe I'm getting this error because the client is sending an empty list for `blocked_tags` but the `tags` field is not present because it is empty and this PR has not been merged. 

You'll note that I'm able to get past the error if I specify valid values for `tags` and `blocked_tags` in the provider, indicating that the client works fine with the `omitempty` struct tag present.

ponbiki:
> > The omitempty needed removed from both unfor
> 
> @ponbiki Looks like your message was incomplete. Why do they both need removed? Are `tags` and `blocked_tags` now required fields of the API? The documentation does not seem to indicate that.
> 
> https://ns1.com/api?docId=2185
> 
> The removal of the `omitempty` struct tag is causing your client to send the `blocked_tags` field to the API even when it has no content now where it was not sent before, resulting in the error I reported:
> 
> ```
> Error: POST https://api.nsone.net/v1/zones/MYZONE/MYRECORD/A: 400 tags and blocked_tags must be updated together
> ```
> 
> I believe I'm getting this error because the client is sending an empty list for `blocked_tags` but the `tags` field is not present because it is empty and this PR has not been merged.
> 
> You'll note that I'm able to get past the error if I specify valid values for `tags` and `blocked_tags` in the provider, indicating that the client works fine with the `omitempty` struct tag present.

My apologies for the unintended comment post.  The NS1 API requires that if `tags` or `blocked_tags` is POSTed, then the other must also always be POSTed even if said value is `NULL`/`{}`/`[]`.  The `omitempty` was meant to be removed from both `tags` and `blocked_tags` as it was causing the same 400 error to be raised from the API you are seeing (which is because `blocked_tags: NULL` is being sent without corresponding `tags`.  
Edit: This was not a limitation of the original tags design for the self-hosted product (which I believe is documented above).  The tag behavior is slightly different in the managed product.

pburrows-ns1 (approved):

PR #226, opened 2024-01-29
Add new redirect objects

fformica:
Adding the new redirect objects (not yet public), and fixing unit tests.
Also, removing deprecated golint, as the functionality should be included in go vet now


rest/redirect.go
```diff
@@ -0,0 +1,174 @@
...
+
+	"gopkg.in/ns1/ns1-go.v2/rest/model/redirect"
+)
+
+// RedirectService handles 'redirect' endpoint.
+type RedirectService service
+
+// List returns the configured redirects.
+//
+// NS1 API docs: TBD
```
fformica: We don't have these docs yet :( Maybe we can add them in a later PR, or delay this one?

dmavrommatis: we will delay the PR

dmavrommatis (dismissed):

dmavrommatis (approved):

PR #231, opened 2024-03-15
INBOX-2702: Fixes #224 by omitting empty tags

### LINKED ISSUE #224: attempting to create records with 2.7.13 returns 500
Hi,

I'm co-maintaining [dnscontrol's](https://github.com/StackExchange/dnscontrol) ns1 support - which uses ns1-go under the hood.

It seems that 2.7.13 that is available as latest via gopkg is getting 500 from NS1's API when attempting to create or update any records. Unfortunately given that it's a 500, I don't get any more info about what's wrong.

Downgrading to 2.7.12 triggers the issue with tags/blocked_tags, so can't really say if that's an issue in 2.7.12 as well. It seems though, that 2.7.11 is working correctly against the API, so it's definitely something post 2.7.11.

Any combination of nil for tags/blocked_tags doesn't seem to help.

Not sure what more info I can provide, but happy to try anything, i can reproduce it every time in my setup.

thanks!

fformica:
This PR sets `tags` and `blocked_tags` as `omitempty`, which fixes the issue in #224 on the client side: however, this has a side effect of not allowing to empty out existing tags in an update, because an empty array is omitted just like null is.

### DIFF
--- a/rest/model/dns/record.go
+++ b/rest/model/dns/record.go
@@ -31,10 +31,10 @@ type Record struct {
 	Regions data.Regions `json:"regions,omitempty"`
 
 	// Contains the key/value tag information associated to the record
-	Tags map[string]string `json:"tags"` // Only relevant for DDI
+	Tags map[string]string `json:"tags,omitempty"` // Only relevant for DDI
 
 	// List of tag key names that should not inherit from the parent zone
-	BlockedTags []string `json:"blocked_tags"` //Only relevant for DDI
+	BlockedTags []string `json:"blocked_tags,omitempty"` //Only relevant for DDI
 
 	// Read-only fields
 	LocalTags []string `json:"local_tags,omitempty"` // Only relevant for DDI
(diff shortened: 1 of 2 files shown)


fformica:
I tested with the following (on prod0) that we can now update tags independently:
```
	client.Records.Delete("fformica.com", "my.test.a.fformica.com", "A")

	record := dns.NewRecord("fformica.com", "my.test.a.fformica.com", "A", map[string]string{"one": "two"}, nil)
	resp, err := client.Records.Create(record)
	require.NoError(t, err)
	require.Equal(t, 200, resp.StatusCode)

	record, resp, err = client.Records.Get("fformica.com", "my.test.a.fformica.com", "A")
	require.NoError(t, err)
	require.Equal(t, 200, resp.StatusCode)
	require.Equal(t, map[string]string{"one": "two"}, record.Tags)

	record.Tags = map[string]string{"three": "four"}
	resp, err = client.Records.Update(record)
	require.NoError(t, err)
	require.Equal(t, 200, resp.StatusCode)

	record, resp, err = client.Records.Get("fformica.com", "my.test.a.fformica.com", "A")
	require.NoError(t, err)
	require.Equal(t, 200, resp.StatusCode)
	require.Equal(t, map[string]string{"three": "four"}, record.Tags)
```

pburrows-ns1 (approved):

PR #233, opened 2024-04-04
feat: add support for listing account activity

tjhop:
This adds a new service to enable listing NS1 account activity. The endpoint supports many url parameters to adjust response behavior (such as changing/limiting how many activity entries are returned, etc), but for this initial approach, the sdk just uses the endpoint directly with the endpoint's normal default values.

I used a small Terraform config to generate 50 dummy DNS records on my NS1 account to generate some account activity and have confirmed that the endpoint operates as expected -- the values I expected are returned, the response is limited to the last 20 activity entries like the endpoint default, etc.

### DIFF
--- a/mockns1/account_activity.go
+++ b/mockns1/account_activity.go
@@ -3,6 +3,7 @@ package mockns1
 import (
 	"net/http"
 
+	api "gopkg.in/ns1/ns1-go.v2/rest"
 	"gopkg.in/ns1/ns1-go.v2/rest/model/account"
 )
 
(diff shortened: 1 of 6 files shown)


mockns1/account_activity.go
```diff
@@ -0,0 +1,19 @@
...
+// function
+func (s *Service) AddActivityListTestCase(
+	requestHeaders, responseHeaders http.Header,
+	response []*account.Activity,
+) error {
+	return s.AddTestCase(
+		http.MethodGet, "/account/activity", http.StatusOK, requestHeaders,
+		responseHeaders, "", response,
+	)
+}
```
Zach-Johnson: I don't think this is referenced anywhere, maybe add a basic test for the GET?

tjhop: Test file has been added, I'm going to add another run to the suite `>1` parameter as well, to help test fix for https://github.com/ns1/ns1-go/pull/234#discussion_r1552288381 as well

tjhop: additional test suite added to test >1 param calls

Zach-Johnson (approved):

PR #234, opened 2024-04-04
feat: add support for setting URL parameters on client requests

tjhop:
Several NS1 API endpoints allow setting URL parameters to alter the behavior of the endpoint. For instance, a `limit` parameter can be set on the `/account/activity` endpoint to control how many events are returned in the response. This would be useful in #233.

This commit adds support for URL parameters by adding a new struct `Param` to the `rest` package, and updating `rest.client.Do()` and `rest.client.DoWithPagination()` to accept a variable amount of params. In order for higher level services based off of the common service to set url params, they will need to pass them through when invoking the client's `Do()` method, like so:

```golang
resp, err := s.client.Do(req, &al, params...)
```

### DIFF
--- a/mockns1/testcase.go
+++ b/mockns1/testcase.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/stretchr/testify/assert"
+	api "gopkg.in/ns1/ns1-go.v2/rest"
 )
 
 type testCase struct {
@@ -29,6 +30,7 @@ func (s *Service) AddTestCase(
 	method, uri string, returnStatus int,
 	requestHeaders, responseHeaders http.Header,
 	requestBody, responseBody interface{},
+	params ...api.Param,
 ) error {
 	s.stopTimer()
 	defer s.startTimer()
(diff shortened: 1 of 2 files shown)


mockns1/testcase.go
```diff
@@ -29,13 +30,19 @@ func (s *Service) AddTestCase(
...
 ) error {
 	s.stopTimer()
 	defer s.startTimer()
 
 	if !strings.HasPrefix(uri, "/v1/") {
 		uri = "/v1/" + uri
 	}
+
+	for _, p := range params {
+		uri = fmt.Sprintf("%s?%s=%s", uri, p.Key, p.Value)
```
Zach-Johnson: this seems wrong, it needs to swap to `&` after the first url parameter

rest/client.go
```diff
@@ -185,10 +185,20 @@ func SetDDIAPI() func(*Client) {
 	return func(c *Client) { c.DDI = true }
 }
 
+type Param struct {
```
Zach-Johnson, line 189: This should have a comment on it.

mockns1/testcase.go
tjhop: ...can you tell I mostly tested with single param requests? :joy: 

Good catch, fixing

Zach-Johnson (approved):

Issue #240, opened 2024-07-31
"feeds" field is missing from the Answer struct


marcsello:
According to the API documentation here: https://developer.ibm.com/apis/catalog/ns1--ibm-ns1-connect-api/api/API--ns1--ibm-ns1-connect-api#getRecord , the answer fields of the record resource has a "feeds" field. However this field is missing from the client implementation's `Answer` struct here: https://github.com/ns1/ns1-go/blob/72c28cb2abb537ade303b5ab2be5c3cc38e88632/rest/model/dns/answer.go#L14-L27

Extending the struct like this:
```go

type AnswerFeed struct {
  Feed   string `json:"feed"`
  Source string `json:"source"`
}

type Answer struct {
  ...
  Feeds []AnswerFeed `json:"feeds"`
}
```

Seems to be working properly for `Records.Get`.

This field would be very useful to have included, because feeds can not be queried only by their ID (that can be get from a metadata feed pointer) the source ID is needed as well, but there's no other way to get this info.

PR #242, opened 2024-08-01
Add feeds field to Answer

marcsello:
This PR adds the missing "Feeds" field to the "Answer" struct as described in this issue: https://github.com/ns1/ns1-go/issues/240

### DIFF
--- a/rest/model/dns/answer.go
+++ b/rest/model/dns/answer.go
@@ -10,11 +10,18 @@ import (
 	"gopkg.in/ns1/ns1-go.v2/rest/model/data"
 )
 
+// AnswerFeed wraps the values of an Answer's "feed" attribute
+type AnswerFeed struct {
+	FeedID   string `json:"feed"`
+	SourceID string `json:"source"`
+}
+
 // Answer wraps the values of a Record's "filters" attribute
 type Answer struct {
 	ID string `json:"id,omitempty"`
 
-	Meta *data.Meta `json:"meta,omitempty"`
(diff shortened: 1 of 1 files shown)


rest/model/dns/answer.go
```diff
@@ -10,11 +10,18 @@ import (
 	"gopkg.in/ns1/ns1-go.v2/rest/model/data"
 )
 
+// AnswerFeed wraps the values of an Answer's "feed" attribute
+type AnswerFeed struct {
+	FeedID   string `json:"feed"`
+	SourceID string `json:"source"`
```
fformica, line 16: Hi; I'm not sure if they can be missing, but just in case, can you make these fields omitempty? 
```suggestion
	FeedID   *string `json:"feed,omitempty"`
	SourceID *string `json:"source,omitempty"`
```

marcsello: Hi, thanks for the feedback!

As far as I know, `omitempty` is only relevant when marshaling a field, and it seems to me that these fields would never get marshaled, they are only present in the API response.

Also what would be the benefit of using pointers here? Empty strings are already invalid for this data type.

fformica: I suppose I'm mostly worried about Terraform, who tends to write what it reads and has a bad habit of adding extra elements in arrays, both of which would result in empty strings being sent in an update; but looking at the code it will just ignore the new field, so I admit it should be fine.

fformica (approved):

PR #244, opened 2024-10-28
Add support for the v1 alerts api.

hhellyer:
This PR adds support for the alerts api.

- Adds code to call the new alerts api.
- Adds testcases.
- Adds example.


rest/model/alerting/alert.go
```diff
@@ -0,0 +1,46 @@
+package alerting
```
apantelopoulosNS1, line 1: oh nice, more models! its not like that we did not have enough already :D

apantelopoulosNS1: not a jab at the PR but rather than the fact that we have a 1000 models!

rest/_examples/alerts.go
```diff
@@ -0,0 +1,132 @@
...
+	api "gopkg.in/ns1/ns1-go.v2/rest"
+	"gopkg.in/ns1/ns1-go.v2/rest/model/alerting"
+	"gopkg.in/ns1/ns1-go.v2/rest/model/dns"
+	"gopkg.in/ns1/ns1-go.v2/rest/model/monitor"
+)
+
+var client *api.Client
+
+// Helper that initializes rest api client from environment variable.
+func init() {
```
apantelopoulosNS1, line 20: Is this called in the example somewhere?

apantelopoulosNS1 (dismissed):

rest/_examples/alerts.go
hhellyer: If I'm 100% honest this is here because I started by copying one of the other examples and wanted it to follow our existing pattern.
It will always be called as go ensures the `init()` function for a file is always run:
https://go.dev/doc/effective_go#init
(The internet seems rather divided on whether it's a bad thing to use which in general but it's probably ok for samples like this.)

rest/_examples/zones.go
```diff
@@ -62,7 +62,7 @@ func main() {
 	}
 
 	// Add an A record with a single static answer.
-	orchidRec := dns.NewRecord(domain, "orchid", "A")
+	orchidRec := dns.NewRecord(domain, "orchid", "A", nil, nil)
```
shane-ns1, line 65: Is `dns.NewRecord()` a call that is part of the Go API? If so, extending it to require additional parameters is a breaking change, meaning we'd have to go from version 2.12.2 to version 3.

If we don't want version 3 (which probably we do not?) and this is part of our Go API, then we should consider adding a new function instead. `dns.NewRecordEx()` or something like that, if we follow Windows-style naming conventions, or `dns.NewRecord2()` if we follow POSIX-style naming convention (I prefer the Windows apporach in this case).

rest/model/alerting/alert.go
```diff
@@ -0,0 +1,46 @@
...
+	ZoneNames       []string        `json:"zone_names"`
+}
+
+var (
+	zoneAlertType   string = "zone"
+	recordAlertType string = "record"
+)
+
+// TODO - Allow zones/notifier lists to be passed as structs?
+//   - Multiple constructors?
```
shane-ns1: I'm not a fan of TODO comments in our code. If there is a TODO file for our Go API, then I guess we can put it there. Otherwise we can make a GitHub issue with this idea.

rest/client.go
```diff
@@ -13,9 +13,9 @@ import (
 )
 
 const (
-	clientVersion = "2.12.2"
+	clientVersion = "2.13.0"
 
-	defaultBase                   = "http://localhost:80"
+	defaultBase                   = "https://api.nsone.net"
```
apantelopoulosNS1, line 18: why did you change this?

apantelopoulosNS1 (dismissed):

rest/alert.go
```diff
@@ -0,0 +1,207 @@
...
+	// Update the alerts fields with data from api (ensure consistent)
+	resp, err := s.client.Do(req, &alert)
+	if err != nil {
+		return resp, err
+	}
+
+	return resp, nil
+}
+
+// Delete immediately deletes an existing alert.
```
shane-ns1: The word "immediately" might make my Spidey-Sense tingle if I was a customer reading this. Why do we say "immediately"? Is there a _non_-immediate way to delete an alert? Do other commands not work immediately? :laughing:

rest/alert_test.go
```diff
@@ -0,0 +1,562 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"gopkg.in/ns1/ns1-go.v2/mockns1"
+	"gopkg.in/ns1/ns1-go.v2/rest/model/alerting"
+
+	api "gopkg.in/ns1/ns1-go.v2/rest"
+)
+
+func strPtr(str string) *string {
```
shane-ns1, line 15: Annoying that such a function is needed. I should add this to my "Why I Hate Go" presentation!

rest/alert_test.go
```diff
@@ -0,0 +1,562 @@
...
+			// Pass by value here. alertToCreate will be modified by client.Alerts.Create
+			require.Nil(t, mock.AddAlertCreateTestCase(nil, nil, alertToCreate, *alertResponse))
+
+			_, err := client.Alerts.Create(&alertToCreate)
+			require.Nil(t, err)
+			// alerttoCreate will be modified by the call to Create and fully populated.
+			compareAlerts(t, alertResponse, &alertToCreate)
+		})
+
+		// TODO Add test for error response when updating fields that can't be changed.
```
shane-ns1: Is this difficult? Why isn't it in the PR?

shane-ns1 (changes requested):
I have a few questions, nothing very critical.

I am concerned about possibly changing the API without bumping the major version number, but I maybe this concern is misplaced.

rest/_examples/zones.go
hhellyer: Ah - I should have put a note in the PR.
I only changed this because dns.NewRecord has taken 5 params since https://github.com/ns1/ns1-go/pull/220 some time ago and the example was already broken.
I probably should have raised a separate issue. I can move this to a different issue if you like.

rest/alert.go
hhellyer: I copied that from the notifier list sdk code - I was trying to blend in. Removed.

rest/alert_test.go
hhellyer: Removed, it's not difficult it's also not an issue on alert creation - I copied this from either update or replace which does test this.

rest/_examples/zones.go
shane-ns1: No worries. We should have made a major version bump or added a new "constructor" at that time, but we are where we are. :laughing:

Since this is just an example fix from your PR, I think we're good.

shane-ns1 (approved):
Looks good. Thanks for the attention to detail. :+1:

PR #249, opened 2025-04-08
INBOX-3419 - Add support for primary_network on secondary zones

pburrows-ns1:
1. Added primary_network field to secondary zones
2. Fixed heading levels and date formats

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,15 @@
+## 2.14.2 (April 8th, 2025)
+FEATURES:
 
-### 2.14.1 (April 4th, 2025)
+* Adds support for `primary_network` for secondary zones
+
+## 2.14.1 (April 4th, 2025)
 
 BUG FIXES:
 
 * Fixing billing usage models to use `int64` instead of `int` for all usage values
 
-# 2.14.0 (March 31st, 2025)
+## 2.14.0 (March 31st, 2025)
 
 FEATURES:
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,11 +1,15 @@
+## 2.14.2 (April 8th, 2025)
+FEATURES:
+
+* Adds support for `primary_network` for secondary zones
 
-### 2.14.1 (April 4th, 2025)
+## 2.14.1 (April 4th, 2025)
```
fformica, line 6: I love the proofreading 😄

rest/model/dns/zone.go
```diff
@@ -98,9 +98,10 @@ type ZoneSecondary struct {
 	Status  string  `json:"status,omitempty"`
 	Error   *string `json:"error"`
 
-	PrimaryIP   string `json:"primary_ip,omitempty"`
-	PrimaryPort int    `json:"primary_port,omitempty"`
-	Enabled     bool   `json:"enabled"`
+	PrimaryIP      string `json:"primary_ip,omitempty"`
+	PrimaryPort    int    `json:"primary_port,omitempty"`
+	PrimaryNetwork int    `json:"primary_network"`
```
fformica, line 103: I think it should be `omitempty` like `PrimaryIP`

pburrows-ns1: I should have added more context.

When this field is missing from the payload, the API will choose the first element from the list, which may not be the managed network and that is confusing, especially since the networks list is a set in terraform.

If we keep default values, then it will default to managed unless explicitly overwritten by this field.

fformica: Ok, that might catch someone who only has access to a private network by surprise, but on balance it's probably an easier problem to deal with.

fformica (approved):

PR #254, opened 2025-08-08
fix(search): fix search record results

### LINKED ISSUE #251: Record Search does not marshall the correct type
When running a record search (`Client.RecordSearch.Search("domain=something.com")`), it assumes that the API returns a list of Records. 
The REST API does not return Record JSON with the same fields as normal Record lookup calls. 

For example:
### An api request for a particular record using the zone URI:
```
 curl -X GET -H "X-NSONE-Key: $nsone_key" https://api.nsone.net/v1/zones/<zone>/<fqdn>/CNAME
```
will return
```json
{
  "answers": [
    {
      "answer": [
        "somewhere.com"
      ],
      "id": "<id>"
    }
  ],
  "created_at": 1692284943,
  "domain": "<fqdn>",
  "filters": [],
  "meta": {},
  "networks": [ 0 ],
  "regions": {},
  "tier": 1,
  "ttl": 3600,
  "type": "CNAME",
  "updated_at": 1738793446,
  "use_client_subnet": true,
  "zone": "<zone>",
  "zone_name": "<zone>",
  "customer": 999
  "feeds": [],
  "id": "<id>"
}
```
This matches what a Record struct expects.

### A call to the Search endpoint returns a JSON array of records in a different format:
```
curl -X GET -H "X-NSONE-Key: $nsone_key" https://api.nsone.net/v1/dns/record/search\?domain=<domain>
```
which has limited data, and different zone field names:
```json
{
  "limit": 50,
  "next": "",
  "results": [
    {
      "answers": [
        {
          "answer": [
            "somewhere.com"
          ]
        }
      ],
      "domain": "<domain>",
      "ttl": 3600,
      "type": "CNAME",
      "zone_fqdn": "<zone>",
      "zone_handle": "<zone>"
    }
  ],
  "total_results": 1
}
```

### Problem
The [Search function](https://github.com/ns1/ns1-go/blob/v2/rest/search.go#L14) returns type `*dns.SearchResult,`, which makes the assumption that the [list of Results are the same as regular Records](https://github.com/ns1/ns1-go/blob/v2/rest/model/dns/search.go#L7), which is incorrect. [The Records](https://github.com/ns1/ns1-go/blob/v2/rest/model/dns/record.go#L13) struct is not what the search API returns, so search for records with this go library results in objects with very little and not useful data:
```golang
possibleRecords, _, _ := Client.RecordSearch.Search(fmt.Sprintf("domain=%s", domain))
logger.Debug().Msgf("%+v", possibleRecords)
// prints  &{Next: Limit:50 TotalResults:1 Results:[<domain> CNAME]}
```

And since the body of the `http.Response` is closed after it's read, I'm unable to parse the raw JSON and marshal it into correct objects.

### Possible solutions
1. Create a new, correct type for search result records
2. Expand the current Record format to allow marshal both responses correctly (zone and search)
3. Have NS1 update their API with consistent responses (outside the scope of this library)

arnavbhattacharya-NS1:
Fixes https://github.com/ns1/ns1-go/issues/251

### DIFF
--- a/rest/model/dns/search.go
+++ b/rest/model/dns/search.go
@@ -2,14 +2,14 @@ package dns
 
 import "encoding/json"
 
-type SearchResult struct {
-	Next         string          `json:"next"`
-	Limit        int             `json:"limit"`
-	TotalResults int             `json:"total_results"`
-	Results      []*SearchRecord `json:"results"`
+type ZoneSearchResult struct {
+	Next         string        `json:"next"`
+	Limit        int           `json:"limit"`
+	TotalResults int           `json:"total_results"`
+	Results      []*ZoneSearch `json:"results"`
 }
 
-type SearchRecord struct {
+type ZoneSearch struct {
 	Domain     string    `json:"domain"`
 	Type       string    `json:"type"`
 	TTL        int       `json:"ttl"`
@@ -19,14 +19,14 @@ type SearchRecord struct {
 }
 
 // New correct types
-type SearchResultV2 struct {
-	Next         string            `json:"next"`
-	Limit        int               `json:"limit"`
-	TotalResults int               `json:"total_results"`
-	Results      []*SearchRecordV2 `json:"results"`
+type RecordSearchResult struct {
+	Next         string          `json:"next"`
+	Limit        int             `json:"limit"`
+	TotalResults int             `json:"total_results"`
+	Results      []*RecordSearch `json:"results"`
 }
 
-type SearchRecordV2 struct {
+type RecordSearch struct {
 	Domain     string          `json:"domain"`
 	Type       string          `json:"type"`
 	TTL        int             `json:"ttl"`
@@ -36,5 +36,7 @@ type SearchRecordV2 struct {
 }
 
 type SearchAnswer struct {
-	Answer json.RawMessage `json:"answer"`
+	// Answer json.RawMessage `json:"answer"`
+	Raw   json.RawMessage `json:"answer"`
+	Rdata []string        `json:"-"`
 }
--- a/rest/model/dns/search.go
+++ b/rest/model/dns/search.go
@@ -1,8 +1,17 @@
 package dns
 
 type SearchResult struct {
-	Next         string    `json:"next"`
-	Limit        int       `json:"limit"`
-	TotalResults int       `json:"total_results"`
-	Results      []*Record `json:"results"`
+	Next         string          `json:"next"`
+	Limit        int             `json:"limit"`
+	TotalResults int             `json:"total_results"`
+	Results      []*SearchRecord `json:"results"`
+}
+
+type SearchRecord struct {
+	Domain     string    `json:"domain"`
+	Type       string    `json:"type"`
+	TTL        int       `json:"ttl"`
+	ZoneFQDN   string    `json:"zone_fqdn"`
+	ZoneHandle string    `json:"zone_handle"`
+	Answers    []*Answer `json:"answers"`
 }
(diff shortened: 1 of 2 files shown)


rest/search.go
```diff
@@ -32,19 +32,19 @@ func (s *RecordSearchService) Search(params string) (*dns.SearchResult, *http.Re
 type ZoneSearchService service
 
 // Find takes query parameters and returns matching DNS zones.
-func (s *ZoneSearchService) Search(params string) (*dns.SearchResult, *http.Response, error) {
+func (s *ZoneSearchService) Search(params string) ([]*dns.Zone, *http.Response, error) {
```
fformica: Is it really returning a Zone? I tried and all I got was fqdn and handle :/

rest/model/dns/search.go
```diff
@@ -1,8 +1,40 @@
...
+	Domain     string          `json:"domain"`
+	Type       string          `json:"type"`
+	TTL        int             `json:"ttl"`
+	ZoneFQDN   string          `json:"zone_fqdn"`
+	ZoneHandle string          `json:"zone_handle"`
+	Answers    []*SearchAnswer `json:"answers"`
+}
+
+type SearchAnswer struct {
+	Answer json.RawMessage `json:"answer"`
```
fformica: I tried this test and it worked
```
	recs, resp, err := client.RecordSearch.Search("type=A")
	require.NoError(t, err)
	require.Equal(t, 200, resp.StatusCode)
	require.Greater(t, recs.TotalResults, 1)
	require.Equal(t, "A", recs.Results[0].Type)
	require.Equal(t, 3600, recs.Results[0].TTL)
	require.Equal(t, "yellow-1111.demo.delivery", recs.Results[0].Domain)
	require.Equal(t, "1.1.1.1", recs.Results[0].Answers[0].Rdata[0])
```
So to avoid disruptions I would like to keep it working in the new version; the zone search failed to get any result at all so it might be fine to break that.  I think in order to achieve that we need
``` suggestion
	Rdata []string `json:"answer"`
```
I tried with SSHFP, CERT and DNSKEY and I could see the numbers as strings, so this seems safe; do you have an example where this breaks?

rest/model/dns/search.go
```diff
@@ -1,8 +1,40 @@
...
+	Domain     string    `json:"domain"`
+	Type       string    `json:"type"`
+	TTL        int       `json:"ttl"`
+	ZoneFQDN   string    `json:"zone_fqdn"`
+	ZoneHandle string    `json:"zone_handle"`
+	Answers    []*Answer `json:"answers"`
+}
+
+// New correct types
+type SearchResultV2 struct {
```
fformica: Rather than SearchResult and SearchResultV2 I would have RecordSearchResult and ZoneSearchResult; there's no one else using SearchResult, is there?

fformica (changes requested):

rest/model/dns/search.go
fformica: Nevermind I found a CERT with numbers showing as numbers :/ 
```
json: cannot unmarshal number into Go struct field SearchAnswer.results.answers.answer of type string
```
If anyone is using this method there might be no way to make it work properly without breaking what works :/

fformica: What we could do is 
```
type SearchAnswer struct {
	raw json.RawMessage `json:"answer"`
	Rdata []string `json:"-"`
}
```
And then unpacking and converting `raw` into `Rdata` in the method

fformica (approved):
LGTM

PR #255, opened 2025-08-19
usage alerts (account)

haydencarroll-NS1:


rest/usage_alerts.go
```diff
@@ -0,0 +1,205 @@
...
+	if alert.Subtype == nil {
+		return nil, resp, fmt.Errorf("alert is missing subtype")
+	}
+	
+	if _, ok := alerting.AllowedUsageSubtypes[*alert.Subtype]; !ok {
+		return nil, resp, fmt.Errorf("alert is not a usage alert (invalid subtype)")
+	}
+	
+	return alert, resp, nil
+}
```
fformica: Do we really need completely separate methods for usage alerts :/ I think we should reuse the existing alerts as much as possible

CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
...
+
+ENHANCEMENTS:
+
+* Refactored Usage Alerts to use thin helpers that leverage the existing alert functionality
+
+## 2.15.0 (August 19th, 2025)
+
+FEATURES:
+
+* Add Usage Alerts (account scope) with CRUD, list, and client-side validation
```
fformica, line 11: This PR isn't released yet so we only need one entry; we can sort it out after we merge all the PRs though

rest/usage_alerts.go
```diff
@@ -0,0 +1,104 @@
...
+	if err := alerting.ValidateUsageAlert(alert); err != nil {
+		return nil, err
+	}
+	
+	return alert, nil
+}
+
+// UsageAlertPatch defines the fields that can be updated on a usage alert
+// Intentionally excluding Type and Subtype which cannot be changed via PATCH
+type UsageAlertPatch struct {
```
fformica, line 22: TBH this could have been generalised for all alerts, but I'm not sure how useful it is; a user would normally create or read an object and then modify fields in place before updating

fformica (approved):
Looks good overall; just a couple of comments but we can merge anyway

PR #256, opened 2025-09-16
honor explicit empty networks by adding Networks *[]int (deprecate NetworkIDs)

haydencarroll-NS1:


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 2.16.0 (Unreleased)
+
+FEATURES:
```
fformica: I'd call it a bugfix TBH

rest/_examples/networks_empty.go
```diff
@@ -0,0 +1,57 @@
...
+	zone4 := dns.NewZone("legacy.example.com")
+	zone4.NetworkIDs = []int{3, 4}
+	zone4.EnsureNetworksFromLegacy()
+	data4, _ := json.Marshal(zone4)
+	fmt.Println("Case 4 - Legacy NetworkIDs with EnsureNetworksFromLegacy:")
+	fmt.Printf("  JSON: %s\n", string(data4))
+	fmt.Printf("  Contains networks: %v\n", contains(string(data4), `"networks":[3,4]`))
+}
+
+func contains(s, substr string) bool {
```
fformica: Isn't this just `strings.Contains()`?

rest/model/dns/zone.go
```diff
@@ -51,6 +55,41 @@ type Zone struct {
 	Tags map[string]string `json:"tags,omitempty"` // Only relevant for DDI
 }
 
+// UnmarshalJSON ensures backward compatibility by populating NetworkIDs from Networks
+func (z *Zone) UnmarshalJSON(data []byte) error {
```
fformica: I don't like leaving it to the user to call this function.
IMO the new field should be internal (lowercase) and the old field should always win: the equivalent of this function would then be in `func (s *ZonesService) Get()`

fformica:
This should be redundant now after Arnav's PR https://github.com/ns1/ns1-go/pull/257

PR #257, opened 2025-09-30
Fix/zone networks empty array

arnavbhattacharya-NS1:
fix for when a zone is created with the `networks=[]`. The created zone was being attached Network 0(as it gets added to it by default when no `networks` field is used)

### DIFF
--- a/rest/model/dns/zone.go
+++ b/rest/model/dns/zone.go
@@ -35,9 +35,10 @@ type Zone struct {
 
 	// Networks contains the network ids the zone is available. Most zones
 	// will be in the NSONE Global Network(which is id 0).
-	NetworkIDs []int         `json:"-"`
-	Networks   *[]int        `json:"networks"`
-	Records    []*ZoneRecord `json:"records,omitempty"`
+	NetworkIDs []int `json:"-"`
+	// Networks is an Auxiliary field to help with JSON marshalling/unmarshalling of NetworkIDs for internal use only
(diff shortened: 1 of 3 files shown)


rest/model/dns/zone.go
```diff
@@ -51,6 +53,46 @@ type Zone struct {
...
+		Alias: (*Alias)(&z),
+	}
+
+	if z.NetworkIDs != nil {
+		aux.Networks = &z.NetworkIDs
+	}
+
+	if aux.Networks == nil {
+		aux.Networks = &[]int{}
+	}
```
fformica: I think we shouldn't make assumptions on the value of `Networks`, and just force it to nil if `NetworkIDs` is nil

fformica (commented):
It worked nicely in my test:
```
create nil -> []
update [0] -> [0]
update nil -> [0]
update [] -> []
```
Just a comment

rest/model/dns/zone.go
fformica: Interestingly if we do that the first line becomes
```
create nil -> [0]
```
Which is the same behaviour we had with NetworksId (it's intended)

fformica: @arnavbhattacharya-NS1 can you do this?
```suggestion
} else {
	aux.Networks = nil
}
```

fformica (dismissed):
LGTM

fformica (approved):

PR #260, opened 2025-12-19
Zonefile export

ddevine-NS1:
Adds support for zonefile export in sdk; along with examples. 

Some unrelated files were changed after running `gofmt`.


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 2.17.0 (December 19th, 2025)
```
shane-ns1, line 1: :see_no_evil:

rest/_examples/zonefile_export.go
```diff
@@ -0,0 +1,142 @@
...
+	httpClient := &http.Client{Timeout: time.Second * 10}
+	// Adds logging to each http request.
+	doer := api.Decorate(httpClient, api.Logging(log.New(os.Stdout, "", log.LstdFlags)))
+	client = api.NewClient(doer, api.SetAPIKey(k))
+}
+
+func main() {
+	// Step 1: Initiate the zone file export
+	fmt.Printf("Initiating zone file export for %s...\n", zoneName)
+	status, _, err := client.Zones.ExportZonefile(zoneName)
```
shane-ns1: Why don't we grab and print the response too?

rest/_examples/zonefile_export.go
```diff
@@ -0,0 +1,142 @@
...
+		fmt.Printf("Attempt %d/%d - Status: %s", i+1, maxAttempts, status.Status)
+		if status.Message != "" {
+			fmt.Printf(" - %s", status.Message)
+		}
+		fmt.Println()
+
+		switch status.Status {
+		case "COMPLETED":
+			fmt.Printf("Export completed at: %s\n", status.GeneratedAt)
+			goto download
```
shane-ns1: I was taught that GOTO is harmful:

https://homepages.cwi.nl/~storm/teaching/reader/Dijkstra68.pdf

With this kind of loop+switch, I'd probably change the loop invariant, and structure it something like this:
```go
attempts := 0
isCompleted := false
for !isCompleted {
    attempts += 1
    if attempts > maxAttempts {
    	log.Fatal("Export did not complete within the expected time")
    }
    ...
    switch status.Status {
    case "COMPLETED":
		fmt.Printf("Export completed at: %s\n", status.GeneratedAt)
		isCompleted = true
		...
} 
```

rest/_examples/zonefile_export.go
```diff
@@ -0,0 +1,142 @@
...
+//
+// This example shows the complete workflow for exporting a zone file:
+//  1. Initiate the zone file export
+//  2. Poll for export completion status
+//  3. Download the completed zone file
+//
+// Usage:
+//
+//	export NS1_APIKEY="your-api-key-here"
+//	export NS1_ZONENAME="example.com"
```
shane-ns1: I'd usually pass the zone name as a command-line argument, but if this matches other examples then it's fine...

rest/_examples/zonefile_export.go
```diff
@@ -0,0 +1,142 @@
...
+
+	_, err = file.Write(buf.Bytes())
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	fmt.Printf("Zone file saved to: %s\n", filename)
+	fmt.Printf("File size: %d bytes\n", buf.Len())
+
+	// Optionally, print the first few lines of the zone file
```
shane-ns1: Optionally? We always do this right?

rest/zone.go
```diff
@@ -187,6 +188,78 @@ func (s *ZonesService) nextRecords(v *interface{}, uri string) (*http.Response,
...
+			return nil, resp, ErrZoneMissing
+		}
+		return nil, resp, err
+	}
+
+	return &status, resp, nil
+}
+
+// DownloadZonefile downloads the generated zone file for the specified zone.
+// Returns a bytes.Buffer containing the zone file contents.
```
shane-ns1: It might be nice to also provide a streaming version of this, to help download big files. Although honestly we don't have many of those... :thinking:

shane-ns1 (commented):
Overall this looks nice, but of course I have a few questions and suggestions. :sweat_smile:

rest/_examples/zonefile_export.go
ddevine-NS1: https://github.com/ns1/ns1-go/pull/260/commits/b913aec8248c94dca0234c98abdee5dbc29406c5

rest/_examples/zonefile_export.go
ddevine-NS1: Fixed: https://github.com/ns1/ns1-go/pull/260/commits/9196341397bd7e9c3d52168c254aca8c675677a8

Seemed cleaner to just remove the `switch` imo.

rest/_examples/zonefile_export.go
ddevine-NS1: Good idea! https://github.com/ns1/ns1-go/pull/260/commits/e268ef8b68902a3fa81737c98d1f7938331a552a

rest/_examples/zonefile_export.go
ddevine-NS1: Removed that! Yeah I was originally adding a flag for it but seemed a bit over complicated for an example.

rest/zone.go
ddevine-NS1: Steaming version; https://github.com/ns1/ns1-go/pull/260/commits/4feddd38f8e942da28c80a34b3a0e8272caec005

Needed a small change to the client: https://github.com/ns1/ns1-go/blob/7472feebfd43894b09c34335f64619d3a556ccce/rest/client.go

shane-ns1: Nice!

rest/_examples/zonefile_export.go
```diff
@@ -4,139 +4,199 @@
...
 }
 
 func main() {
-	// Step 1: Initiate the zone file export
-	fmt.Printf("Initiating zone file export for %s...\n", zoneName)
-	status, _, err := client.Zones.ExportZonefile(zoneName)
+	useStream := flag.Bool("stream", false, "Use streaming mode to download directly to disk (more memory efficient)")
+	flag.Parse()
+
+	if flag.NArg() < 1 {
```
shane-ns1:
```suggestion
	if flag.NArg() != 1 {
```

Otherwise we'll silently ignore additional arguments.

rest/_examples/zonefile_export.go
```diff
@@ -4,139 +4,199 @@
...
+		log.Fatal("Usage: go run zonefile_export.go [-stream] <zone-name>")
+	}
+
+	zoneName := flag.Arg(0)
+
+	log.Printf("Initiating zone file export for %s...\n", zoneName)
+	exportStatus, resp, err := client.Zones.ExportZonefile(zoneName)
+
+	jsonBytes, _ := json.MarshalIndent(exportStatus, "", "  ")
```
shane-ns1: I'd go ahead and check the error return, just in case:
```suggestion
	jsonBytes, err := json.MarshalIndent(exportStatus, "", "  ")
	if err != nil {
	    log.Fatalf("Error processing JSON response: %v", err)
	}
```

rest/_examples/zonefile_export.go
```diff
@@ -4,139 +4,199 @@
...
-			log.Fatalf("Export failed: %s", status.Message)
-		case "QUEUED", "GENERATING":
+		log.Print(msg)
+
+		if exportStatus.Status == "COMPLETED" {
+			log.Printf("Export completed at: %s\n", exportStatus.GeneratedAt)
+			break
+		}
+		if exportStatus.Status == "FAILED" {
+			log.Fatalf("Export failed: %s", exportStatus.Message)
```
shane-ns1, line 212: We should `break` here too, since FAILED means that it is done.

rest/_examples/zonefile_export.go
```diff
@@ -4,139 +4,199 @@
...
+		}
+		if slices.Contains([]string{"QUEUED", "GENERATING"}, exportStatus.Status) {
 			// Continue polling
 			time.Sleep(pollInterval)
-		default:
-			log.Fatalf("Unknown status: %s", status.Status)
+			continue
 		}
+		log.Fatalf("Unknown exportStatus: %s", exportStatus.Status)
```
shane-ns1, line 219: We should either invoke `time.Sleep()` on unknown errors or exit with error, otherwise we will busy-loop through all of our attempts repeating the same `"Unknown exportStatus"` message, and then fail.

rest/zone.go
shane-ns1: Actually, looking at this more, the changes are good because they save one buffer copy, but they still read the entire answer into memory first.

I found a Stack Overflow question which seems to have the "proper" way to handle this:

https://stackoverflow.com/a/22177737

I don't insist on this approach - we can leave it as a ticket if we want - but if you feel like it then it is the complete fix. :smile:

shane-ns1 (changes requested):
I have a few mostly-minor suggestions, but also an idea about the streaming version. :see_no_evil:

rest/_examples/zonefile_export.go
ddevine-NS1: `log.Fatalf` calls `os.Exit(1)` so it wouldn't reach the break

rest/_examples/zonefile_export.go
ddevine-NS1: Same here: `log.Fatalf` calls `os.Exit(1)`;

rest/zone.go
ddevine-NS1: Good idea! https://github.com/ns1/ns1-go/pull/260/commits/e1cdadfa1c898ac1ba8d04a8e907224f9fb0dcef

So there's now three ways to download the file

1. `DownloadZonefileWriter` -  takes a `io.writer` object; this is handy for "automatic" handling, so someone can just pass a `File` object and it should just stream to disk for them. 
2. `DownloadZonefile` - Basic handling, returns a bytes.Buffer object with the zonefile in memory - this is the way other parts of the sdk seem to work, so left it here for consistency. 
3. `DownloadZonefileReader` - Returns a reader like in the example and can process it line by line. 

I updated [the example file](https://github.com/ns1/ns1-go/blob/a7ed9d766d35c6bbaa35b0b96779b67a739c8554/rest/_examples/zonefile_export.go) to use each of them based on a flag.

rest/_examples/zonefile_export.go
```diff
@@ -45,8 +47,121 @@ func init() {
...
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close() // IMPORTANT: Must close when done
+
+	noLines := 0
+
+	for {
+		line, err := reader.ReadString('\n')
+		if err == io.EOF {
```
shane-ns1, line 72: https://pkg.go.dev/bufio#Reader.ReadString

>  If ReadString encounters an error before finding a delimiter, it returns the data read before the error and the error itself (often io.EOF). 

This shouldn't be a problem IFF we always have a newline on our zone files (which I guess we do). Otherwise we would have to save the EOF signal and handle it after we've processed `line`. :stuck_out_tongue_closed_eyes:

shane-ns1 (commented):
Thanks for adding a "true" streaming mode. Is it possible to use the `DownloadZonefileReader()` technique for the `DownloadZonefileWriter()` function as well? Otherwise the `DownloadZonefileWriter()` will still load the whole zone into memory before writing it to disk...

shane-ns1 (approved):

shane-ns1:
> Thanks for adding a "true" streaming mode. Is it possible to use the `DownloadZonefileReader()` technique for the `DownloadZonefileWriter()` function as well? Otherwise the `DownloadZonefileWriter()` will still load the whole zone into memory before writing it to disk...

It turns out I misunderstood the `net.http` interface, and we are already streaming in all cases. So... yay!

PR #266, opened 2026-04-03
Peng 7928/golang sdk apikey expiry

soniafrancisNS1:
## Summary
Implements Go SDK support for API key rotation feature.
https://ns1inc.atlassian.net/browse/PENG-7928
## Changes
- Add rotation data structures to APIKey model
- Add UpdateSecret and DeleteSecret methods
- Add tests for secret management methods

## Related
https://ns1inc.atlassian.net/browse/PENG-8192
https://ns1inc.atlassian.net/browse/PENG-8193


rest/account_apikey.go
```diff
@@ -142,9 +142,60 @@ func (s *APIKeysService) Delete(keyID string) (*http.Response, error) {
 	return resp, nil
 }
 
+// UpdateSecret updates an API key secret's enabled status or expiration date.
+
+func (s *APIKeysService) UpdateSecret(secretID string, edit *account.APIKeySecretEdit) (*account.APIKeySecret, *http.Response, error) {
```
hhellyer: I think looking at other implementations that the signature of this function is wrong.
It should be:
```
func (s *APIKeysService) UpdateSecret(secret *account.APIKeySecret) (*http.Response, error)
```
The way other update functions seem to work is that the object id to update is also contained in the `secret` parameter.
There is no need for a separate return object as the `Do` method below will update the object it is passed.

It's a bit confusing but the pattern makes more sense if you think about how the Create or Get functions are used in code.
You perform a create or a get and an `account.APIKeySecret` pointer is returned. Then you change any fields you want to modify and call `UpdateSecret`. That way there's no need for a `APIKeySecretEdit` structure.

soniafrancisNS1: The method now takes `_secret *account.APIKeySecret_` as the parameter and returns `(_*http.Response, error)._ `The `_object ID_` is contained in the secret parameter itself, and the `_Do method_` updates the object in place, so there's no need for a separate return object. Also removed the `_APIKeySecretEdit_` struct entirely.

rest/_examples/apikey_rotation.go
```diff
@@ -0,0 +1,243 @@
...
+		newSecret, _, err := client.APIKeys.RenewSecret(secretID)
+		if err != nil {
+			log.Fatal(err)
+		}
+		fmt.Printf("New Secret Created:\n")
+		fmt.Printf("  Secret ID: %s\n", newSecret.ID)
+		fmt.Printf("  Secret Key: %s\n", newSecret.Key) // Only visible once!
+		fmt.Printf("  Expires At: %s\n", newSecret.ExpiresAt)
+		fmt.Printf("  Enabled: %t\n", *newSecret.Enabled)
+		fmt.Println("\n⚠️  IMPORTANT: Save the secret key immediately - it won't be shown again!")
```
hhellyer: It might be worth removing the warning triangle symbol. I'm not sure how well it will move between systems.

rest/_examples/apikey_rotation.go
```diff
@@ -0,0 +1,243 @@
...
+
+	// Example 10: Self-service operations (using the API key itself)
+	// Note: These would typically be called by the API key itself, not an admin
+	fmt.Println("\n=== Self-Service Operations ===")
+	fmt.Println("(These operations allow an API key to manage itself)")
+
+	// Get own secret details
+	selfSecret, _, err := client.APIKeys.GetSecretSelf()
+	if err != nil {
+		// This might fail if not using the actual secret for auth
```
hhellyer: If this is going to fail by default when the example is run could you create a new `client` object that uses the new secret value so that it works?

rest/account_apikey.go
```diff
@@ -142,9 +142,175 @@ func (s *APIKeysService) Delete(keyID string) (*http.Response, error) {
...
+
+// RenewSecretSelf renews the API key secret used in the current request.
+// This generates a new secret value with an updated expiration date.
+// This allows an API key to renew itself without needing manage_apikeys permission.
+// The API key must have an expiry_duration set, and cannot have more than 2 active secrets.
+// Returns the new secret with the plaintext secret value (only time it's visible).
+//
+// NS1 API docs: https://ns1.com/api/#apikeys-v1-secrets-self-renew-post
+func (s *APIKeysService) RenewSecretSelf() (*account.APIKeySecret, *http.Response, error) {
+	path := "apikeys/v1/secrets/self/renew"
```
hhellyer: Could this and the other `self` function just call their equivalent function with a `secretID` of `"self"`?
(Just to save some duplication.)
I think it is worth having the `Get/RenewSecretSelf()` functions - it saves the user having to know the magic string.

hhellyer (changes requested):
This looks good, I've added some minor comments. The example is very comprehensive!
One thing that's missing is updating the version info. There's an example of this in Ferdinando's last PR: https://github.com/ns1/ns1-go/commit/ad12356957fe389def42b18d539fcab76577bd24
I think we should probably update that to 2.18.0 (unless anyone releases before us) and update the changelong but I'm not sure if we want to put it in this PR or a different one when we are ready to release.
We should review how this module gets released - I'm not sure if it's pulled directly from the repo or if there's a release process like there is for Python which uploads a package. If this is live immediately we probably should put the version update and changelog in this PR.

rest/_examples/apikey_rotation.go
soniafrancisNS1: I’ve removed the warning triangle symbol from the example output to keep it cleaner across systems.

soniafrancisNS1: Done, removed the warning triangle symbol.

rest/account_apikey.go
soniafrancisNS1: `GetSecretSelf() `now calls `GetSecret("self")` and `RenewSecretSelf()` calls RenewSecret("self"). Also removed the now-unused` ErrInvalidAuth` error variable.

rest/_examples/apikey_rotation.go
soniafrancisNS1: Example 10 now creates a new client using the renewed secret key so it works correctly when run.

hhellyer (dismissed):
LGTM - Thanks for making changes.

hhellyer (approved):
Approving release date update.

